PR #90, opened 2022-08-11
BREAKING: Bump Node to v14

bowensanders:
This changes all refs to v12 to point to v14 as v12 is no longer LTS


.github/workflows/build-test.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x]
+        node-version: [14.x, 16.x]
```
rekmarks, line 14: We should add `18.x` to this array as well.

rekmarks (changes requested):

.github/workflows/build-test.yml
bowensanders: Requesting permission to wait to add 18.x support to this PR (and the few others) until after vacation.

FrederikBolding: Done

Mrtenz (changes requested):
Please update `README.md` too.

Mrtenz (approved):

PR #95, opened 2022-09-08
Replace `ethers` with submodules

adonesky1:
This is https://github.com/MetaMask/metamask-extension/pull/15461 to remove any ethers submodules that are not used in the extension dependency tree by replacing all imports of ethers for the used submodule itself

CHANGED:

- Simply replaces the "umbrella" version of ethers with the submodules that we actually use: @ethersproject/bytes , @ethersproject/providers & @ethersproject/bignumber.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,10 +26,12 @@
     "build:link": "yarn build && cd dist && yarn link && rm -rf node_modules && cd .."
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


darkwing (approved):
Excellent work @adonesky1 !

dan437 (approved):
It's working, LGTM.

PR #110, opened 2022-11-24
Use controllers monorepo

FrederikBolding:
Use new controllers monorepo packages instead of deprecated `@metamask/controllers`.


Gudahtt (approved):
LGTM!

PR #119, opened 2023-02-15
chore: include regular tx events during the stx status changes

nikoferro:
Regular Swaps fire a ‘transaction submitted’ event, but we aren’t currently triggering this event when a Smart Transaction is submitted. 

While the regular tx and an stx go through different flows, we need to map regular tx events to events fired during stx statuses updates, while still keeping STX custom events.

This PR adds `Transaction Submitted` and `Transaction Finalized` (in the event that the STX is cancelled)

Made this flow chart so its easier to understand at which point in time these events are included.

a (1)

### DIFF
--- a/src/SmartTransactionsController.ts
+++ b/src/SmartTransactionsController.ts
@@ -230,7 +230,7 @@ export default class SmartTransactionsController extends BaseController<
     }
 
     const sensitiveProperties = {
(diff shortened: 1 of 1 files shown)


nikoferro:
tagging @forest-diggs-consensys and @John.Brennan for visibility since they have been part of the conversations around this update

PR #123, opened 2023-02-27
Remove `getNetwork` option

mcmire:
Currently NetworkController captures the network ID of the currently selected network via the `net_version` RPC method. However, as it is not part of the Ethereum spec, `net_version` is not guaranteed to be supported across all EVM chains; we can safely use chain ID as a replacement for network ID instead. For this reason, we plan on removing this captured network ID from NetworkController.

The `getNetwork` option to SmartTransactionController was intended to get the network ID as stored by NetworkController, but if we are removing that, then we no longer need this option.

This is a **breaking** change.

---

For more context, see: https://github.com/MetaMask/metamask-extension/issues/16946.

### DIFF
--- a/src/SmartTransactionsController.test.ts
+++ b/src/SmartTransactionsController.test.ts
@@ -6,8 +6,6 @@ import SmartTransactionsController, {
 import { API_BASE_URL, CHAIN_IDS } from './constants';
 import { SmartTransaction, SmartTransactionStatuses } from './types';
 
-const confirmExternalMock = jest.fn();
-
 jest.mock('@ethersproject/bytes', () => ({
   ...jest.requireActual('@ethersproject/bytes'),
   hexlify: (str: string) => `0x${str}`,
@@ -27,6 +25,65 @@ jest.mock('@ethersproject/providers', () => ({
     getBlock = jest.fn();
   },
 }));
+
+/**
+ * The object that the callback to `withController` will be called with.
+ */
+interface WithControllerCallbackArgs {
+  /**
+   * The constructed instance of SmartTransactionsController.
+   */
+  controller: SmartTransactionsController;
+  /**
+   * The function that will effectively be called when the network changes.
+   * You'll want to call this yourself in tests.
+   */
+  networkStateChangeListener: (networkState: NetworkState) => void;
+  /**
+   * The current date, as stubbed via `jest.useFakeTimers()`.
+   */
+  now: Date;
+}
+
+/**
+ * The callback for `withController`.
+ */
+type WithControllerCallback<ReturnValue> = (
+  args: WithControllerCallbackArgs,
+) => Promise<ReturnValue> | ReturnValue;
+
+/**
+ * The options bag that `withController` takes.
+ */
+interface WithControllerOptions {
+  /**
+   * Will be turned into the first argument that the controller constructor
+   * takes.
+   */
+  options?: Partial<
+    ConstructorParameters<typeof SmartTransactionsController>[0]
+  >;
+  /**
+   * Will be turned into the second argument that the controller constructor
+   * takes.
+   */
+  config?:
+    | ConstructorParameters<typeof SmartTransactionsController>[1]
+    | undefined;
+  /**
+   * Will be turned into the third argument that the controller constructor
+   * takes.
+   */
+  state?:
+    | ConstructorParameters<typeof SmartTransactionsController>[2]
+    | undefined;
+}
+
+type WithControllerArgs<ReturnValue> =
+  | [WithControllerCallback<ReturnValue>]
+  | [WithControllerOptions, WithControllerCallback<ReturnValue>];
+
+const confirmExternalMock = jest.fn();
 
 const addressFrom = '0x268392a24B6b093127E8581eAfbD1DA228bAdAe3';
 
@@ -143,10 +200,6 @@ const createGetFeesApiResponse = () => {
       },
     ],
   };
-};
-
-const createSubmitTransactionsApiResponse = () => {
-  return { uuid: 'dP23W7c2kt4FK9TmXOkz1UM2F20' };
 };
 
 // TODO: How exactly a signed transaction should look like?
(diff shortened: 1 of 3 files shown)


legobeat (approved):
LGTM

Gudahtt (approved):
LGTM!

src/SmartTransactionsController.ts
```diff
@@ -576,7 +570,7 @@ export default class SmartTransactionsController extends BaseController<
       this.updateSmartTransaction({
         chainId,
         nonceDetails,
-        metamaskNetworkId,
+        metamaskNetworkId: chainId.replace(/^0x/u, ''),
```
mcmire, line 573: Oh, whoops, this isn't right. We need to convert the chain id from hex to decimal. I'll fix this.

Gudahtt: Sorry I missed this line. What is this `metamaskNetworkId` used for - do we need it at all?

src/SmartTransactionsController.test.ts
```diff
@@ -450,29 +502,53 @@ describe('SmartTransactionsController', () => {
...
-        signedTransactions: [signedTransaction],
-        signedCanceledTransactions: [signedCanceledTransaction],
-        txParams: signedTransaction,
-      });
-
-      expect(
-        smartTransactionsController.state.smartTransactionsState
-          .smartTransactions[CHAIN_IDS.ETHEREUM][0].uuid,
-      ).toStrictEqual('dP23W7c2kt4FK9TmXOkz1UM2F20');
+      const chainIdHex = '0x64';
```
mcmire, line 506: I updated this test so that it didn't rely on the `beforeEach` above — it creates its own instance of SmartTransactionsController so that we can override the `config` — and also moves the UUID which was previously set in another function into this test so it's easier to see its relationship with the expected value.

src/SmartTransactionsController.ts
mcmire: I'm not 100% sure, but I am assuming that the transaction objects that this controller stores include `metamaskNetworkId` because that property is a part of the TransactionMeta interface in the extension. It's used all over the place there, but in particular the `tx-state-manager` also includes/handles this property, e.g.: https://github.com/MetaMask/metamask-extension/blob/90ae4980b3aca3f2610a24ade06cf7e6db928209/app/scripts/controllers/transactions/tx-state-manager.js#L131 and also https://github.com/MetaMask/metamask-extension/blob/90ae4980b3aca3f2610a24ade06cf7e6db928209/app/scripts/controllers/transactions/tx-state-manager.js#L278. As you can see in the second link, I don't think it's a required property, but I tried to get rid of it as a part of https://github.com/MetaMask/metamask-extension/pull/17556 and wasn't sure whether it would break anything. I would assume the same applies to this transaction. But I'm happy to investigate this.

mcmire:
Okay ready for review again ^

src/SmartTransactionsController.test.ts
```diff
@@ -450,29 +502,53 @@ describe('SmartTransactionsController', () => {
...
+          config: {
+            chainId: chainIdHex,
+            supportedChainIds: [chainIdHex],
+          },
+        },
+        async ({ controller }) => {
+          const signedTransaction = createSignedTransaction();
+          const signedCanceledTransaction = createSignedCanceledTransaction();
+          nock(API_BASE_URL)
+            .post(`/networks/${chainIdDec}/submitTransactions`)
```
legobeat, line 521:
```suggestion
            .post(`/networks/${chainId}/submitTransactions`)
```

src/SmartTransactionsController.test.ts
```diff
@@ -450,29 +502,53 @@ describe('SmartTransactionsController', () => {
...
+          jest.runOnlyPendingTimers();
+          await promise;
+
+          expect(
+            controller.state.smartTransactionsState.smartTransactions,
+          ).toStrictEqual({
+            [chainIdHex]: [
+              expect.objectContaining({
+                chainId: chainIdHex,
+                metamaskNetworkId: chainIdDec.toString(),
```
legobeat, line 538:
```suggestion
                metamaskNetworkId: chainId.toString(),
```

src/SmartTransactionsController.test.ts
```diff
@@ -450,29 +502,53 @@ describe('SmartTransactionsController', () => {
...
+                uuid,
+                history: [
+                  expect.objectContaining({
+                    chainId: chainIdHex,
+                    metamaskNetworkId: chainIdDec.toString(),
```
legobeat, line 543:
```suggestion
                    metamaskNetworkId: chainId.toString(),
```

src/SmartTransactionsController.test.ts
legobeat, line 538:
```suggestion
                metamaskNetworkId: chainId.toString(),
```

src/SmartTransactionsController.test.ts
```diff
@@ -450,29 +502,53 @@ describe('SmartTransactionsController', () => {
...
-        signedCanceledTransactions: [signedCanceledTransaction],
-        txParams: signedTransaction,
-      });
-
-      expect(
-        smartTransactionsController.state.smartTransactionsState
-          .smartTransactions[CHAIN_IDS.ETHEREUM][0].uuid,
-      ).toStrictEqual('dP23W7c2kt4FK9TmXOkz1UM2F20');
+      const chainIdHex = '0x64';
+      const chainIdDec = 100;
```
legobeat, line 507: Nit: `chainIdDec` is a number, not decimal
```suggestion
      const chainId = 100;
      const chainIdHex = `0x${chainId.toString(16)}`; // 0x64
```

src/SmartTransactionsController.ts
Gudahtt: Hmm. It might be safer to try to get rid of it, rather than convert `chainId` to `networkId` like this. They aren't equivalent, so this would lead to transactions on some networks having the wrong `networkId` assigned.

At least it would in theory - I don't think the smart transaction controller is used on any such networks.

mcmire: Okay, I've looked through the extension, mobile, and core codebases, and the only places where `metamaskNetworkId` seems to be used are:

* Transaction activity logging: https://github.com/MetaMask/metamask-extension/blob/987daee85483c38ec9592a05fb8efe1f0e71eda7/ui/components/app/transaction-activity-log/transaction-activity-log.util.js#L101
* Transaction metrics: https://github.com/MetaMask/metamask-extension/blob/987daee85483c38ec9592a05fb8efe1f0e71eda7/app/scripts/controllers/transactions/index.js#L2305
* Removal of old transactions when adding a transaction: https://github.com/MetaMask/metamask-extension/blob/987daee85483c38ec9592a05fb8efe1f0e71eda7/app/scripts/controllers/transactions/tx-state-manager.js#L278
* Optimism: https://github.com/MetaMask/metamask-extension/blob/987daee85483c38ec9592a05fb8efe1f0e71eda7/ui/helpers/utils/optimism/buildUnserializedTransaction.js#L22 (and similar in mobile too)

Except for Optimism it seems like these usages are either insignificant or can be replaced with `chainId` in most cases. Of course there are exceptions to this rule, so it depends on the network. In this case, if you feel confident that smart transactions won't be used on a network that doesn't have a proper chain ID then I can remove this.

But just to make sure, @dan437, do you have any thoughts on this?

src/SmartTransactionsController.test.ts
mcmire: Oh sure that works!

mcmire:
Originally in https://github.com/MetaMask/metamask-extension/issues/16946 I had decided to remove the code from NetworkController that was responsible for fetching and storing the ID of the current network, which would have had repercussions across other controllers, including this one. But I ended up reverting that change, which means that the change to this repo is no longer needed. I'll close this PR for the time being, and we can revive it if and when we decide to remove the concept of network ID completely.

PR #138, opened 2023-05-16
Sync GitHub workflow files with module template

mcmire:
- Instead of running the "build", "lint", and "test" steps in one job, split them up into separate jobs so CI will run faster
- Notify the metamask-npm-publishers group in Slack when a new release is awaiting approval
- Remove the "require additional reviewer" check
- Add a pull request template

Note that the workflows to publish and deploy docs are missing because `typedoc` isn't present in this project; that will be added later.

Also note that in order to ~merge this PR~ make a new release, we will need to add the following secrets:

- `NPM_TOKEN`
- `SLACK_WEBHOOK_URL`


Gudahtt (approved):
LGTM!

Gudahtt:
The `NPM_TOKEN` token is already setup here (this repository already has npm publishing). The slack token is an organization secret, so that's setup already as well.

PR #184, opened 2023-07-26
Use Caip Chain ID

jiexi:
* Use Caip Chain ID

See: https://github.com/MetaMask/utils/pull/116
See: https://github.com/MetaMask/core/pull/1489


jiexi:
This still needs core and util changes and package.json update

jiexi:
Postponed. Not sure if/when will be picked back up.

PR #191, opened 2023-08-29
Remove networkId

jiexi:
Wallets shouldn't be directly concerned about the network ID as this more of a p2p concept for gossip. What wallets really care about is chain ID as that is the correct value to use to identify a chain, build transactions, etc. Although these two values usually match (ignoring hex/dec formatting), there are [exceptions](https://medium.com/@pedrouid/chainid-vs-networkid-how-do-they-differ-on-ethereum-eec2ed41635b).

We want to remove usage of `networkId` from the SmartTransactionController.

* Fixes [mmp-1068](https://github.com/MetaMask/MetaMask-planning/issues/1068)
* See: [core-1633](https://github.com/MetaMask/core/pull/1633)
* See: https://github.com/MetaMask/metamask-extension/pull/20652

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- **BREAKING:** Bump dependency on `@metamask/network-controller` to ^13.0.0 ([#191](https://github.com/MetaMask/smart-transactions-controller/pull/191))
+- Bump dependency on `@metamask/base-controller` to ^3.2.1 ([#191](https://github.com/MetaMask/smart-transactions-controller/pull/191))
+- Bump dependency on `@metamask/controller-utils` to ^5.0.0 ([#191](https://github.com/MetaMask/smart-transactions-controller/pull/191))
+
+### Removed
+- **BREAKING:** Remove  `metamaskNetworkId` from smart transaction state ([#191](https://github.com/MetaMask/smart-transactions-controller/pull/191))
+- Remove `getNetwork` from constructor options ([#191](https://github.com/MetaMask/smart-transactions-controller/pull/191))
 
 ## [4.0.0]
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Changed
-- **BREAKING:** Bump dependency on `@metamask/network-controller` to ^13.0.0 ([#191](https://github.com/MetaMask/smart-transactions-controller/pull/191))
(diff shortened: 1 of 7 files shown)


package.json
jiexi, line 1: reminder to update this after core changes merged

src/SmartTransactionsController.ts
```diff
@@ -575,7 +569,6 @@ export default class SmartTransactionsController extends BaseController<
       this.updateSmartTransaction({
         chainId,
         nonceDetails,
-        metamaskNetworkId,
```
mcmire, line 578: Are we sure that in all places where `metamaskNetworkId` is referred to in the extension, chain ID is used instead?

jiexi: Uncertain how easy it will be to fully transition away from `metamaskNetworkId` for old transactions that predate `chainId`.

See https://github.com/MetaMask/metamask-extension/pull/20652#issuecomment-1707351348

jiexi: This is now resolved on extension side

dan437 (approved):
That worked!

shanejonas (approved):

BelfordZ:
Needs the previews replaced with the releases, but looks good otherwise

jiexi:
Blocked until https://github.com/MetaMask/core/pull/1633 is merged and released

package.json
```diff
@@ -24,6 +24,10 @@
     "test": "jest",
     "test:watch": "jest --watchAll"
   },
+  "resolutions": {
+    "@metamask/controller-utils@^4.1.0": "npm:@metamask-previews/controller-utils@4.3.1-preview.3dbf14f",
+    "@metamask/network-controller@^10.3.0": "npm:@metamask-previews/network-controller@12.1.1-preview.3dbf14f"
```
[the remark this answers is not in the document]
jiexi: https://github.com/MetaMask/core/pull/1633

Bumping NetworkController to v12.2.0 should be fine

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- **BREAKING:** Bump dependency on `@metamask/network-controller` to ^13.0.0 ([#191](https://github.com/MetaMask/smart-transactions-controller/pull/191))
```
adonesky1, line 9: not that it matters since this will be a major dep bump anyways but why is this breaking in the context of this package?

adonesky1 (dismissed):
One non-blocking question

CHANGELOG.md
jiexi: Great catch. I guess technically it isn't breaking because it doesn't directly use the breaking changes from NetworkController. Only thing referenced is the `NetworkState` type, and it doesn't touch `networkId` there. I'll remove the BREAKING

jiexi: Done: https://github.com/MetaMask/smart-transactions-controller/pull/191/commits/85b5b7efbdf8c1431f72ff8f8554f71e7c990a2f

adonesky1 (approved):

PR #210, opened 2023-10-11
Parameterize SmartTransactionsController state by ChainId for MultiChain + Integrate PollingController Mixin

adonesky1:
Resolves: https://github.com/MetaMask/MetaMask-planning/issues/1039

Integrates new `PollingController` abstraction into the `SmartTransactionsController`, implements the `_executePoll` method and parameterizes the `updateSmartTransactions` method by networkClientId such that pending SmartTransaction statuses can be polled for concurrently across different chains.

A note: As is our current controller refactor strategy, these changes are additive, backwards compatible and coexist alongside the existing globally selected network pattern. Once we fully adopt this polling pattern we should no longer access the root `liveness` and `fees` state but rather access these values by chainId from `livenessByChainId` and `feesByChainId`.

### Added
- Integrates `PollingController` mixin into the `SmartTransactionsController` and implements the abstract `_executePoll` method.
### Changed
- Adds optional options object containing a `networkClientId` argument to the `updateSmartTransaction` method which, if passed, is used fetch the chainId that corresponds to the `networkClientId` and is then used to fetch and update the status of pending smart transactions for that chainId.

### DIFF
--- a/src/SmartTransactionsController.test.ts
+++ b/src/SmartTransactionsController.test.ts
@@ -1,7 +1,7 @@
 import nock from 'nock';
 import { NetworkState } from '@metamask/network-controller';
 import { convertHexToDecimal } from '@metamask/controller-utils';
-import { Web3Provider } from '@ethersproject/providers';
+import EthQuery from '@metamask/eth-query';
 import SmartTransactionsController, {
   DEFAULT_INTERVAL,
 } from './SmartTransactionsController';
@@ -24,20 +24,40 @@ jest.mock('@ethersproject/bytes', () => ({
   hexlify: (str: string) => `0x${str}`,
 }));
 
-jest.mock('@ethersproject/providers', () => ({
-  Web3Provider: class FakeProvider {
-    getBalance = () => ({ toHexString: () => '0x1000' });
+jest.mock('@metamask/eth-query', () => {
+  return class FakeEthQuery {
+    sendAsync = jest.fn(({ method }, callback) => {
+      switch (method) {
+        case 'getBalance': {
+          callback(null, { toHexString: () => '0x1000' });
+          break;
+        }
 
-    getTransactionReceipt = jest.fn(() => ({ blockNumber: '123' }));
+        case 'getTransactionReceipt': {
+          callback(null, { blockNumber: '123' });
+          break;
+        }
 
-    getTransaction = jest.fn(() => ({
-      maxFeePerGas: { toHexString: () => '0x123' },
-      maxPriorityFeePerGas: { toHexString: () => '0x123' },
-    }));
+        case 'getBlock': {
+          callback(null, { baseFeePerGas: { toHexString: () => '0x123' } });
+          break;
+        }
 
-    getBlock = jest.fn();
-  },
-}));
+        case 'getTransaction': {
+          callback(null, {
+            maxFeePerGas: { toHexString: () => '0x123' },
+            maxPriorityFeePerGas: { toHexString: () => '0x123' },
+          });
+          break;
+        }
+
+        default: {
+          throw new Error('Invalid method');
+        }
+      }
+    });
+  };
+});
 
 const addressFrom = '0x268392a24B6b093127E8581eAfbD1DA228bAdAe3';
 
@@ -394,7 +414,7 @@ describe('SmartTransactionsController', () => {
 
     it('calls stop if there is a timeoutHandle and no pending transactions', () => {
       const stopSpy = jest.spyOn(smartTransactionsController, 'stop');
-      smartTransactionsController.timeoutHandle = setInterval(() => ({}));
+      smartTransactionsController.timeoutHandle = setTimeout(() => ({}));
       smartTransactionsController.checkPoll(smartTransactionsController.state);
       expect(stopSpy).toHaveBeenCalled();
       clearInterval(smartTransactionsController.timeoutHandle);
@@ -807,16 +827,19 @@ describe('SmartTransactionsController', () => {
         successfulStx as SmartTransaction,
         {
           chainId: CHAIN_IDS.ETHEREUM,
-          ethersProvider: new Web3Provider(jest.fn()),
+          ethQuery: new EthQuery({ sendAsync: jest.fn() }),
         },
       );
       expect(confirmExternalMock).toHaveBeenCalled();
     });
 
     it('throws an error if ethersProvider fails', async () => {
-      smartTransactionsController.ethersProvider.getTransactionReceipt.mockRejectedValueOnce(
-        'random error' as never,
-      );
+      // smartTransactionsController.ethQuery.getTransactionReceipt.mockRejectedValueOnce(
+      //   'random error' as never,
+      // );
+      // jest.spyOn('@metmask/eth-query', 'sendAsync')(() => {
+      //   throw new Error('random error');
+      // });
       const successfulStx = {
         ...createStateAfterSuccess()[0],
         history: testHistory,
@@ -825,7 +848,7 @@ describe('SmartTransactionsController', () => {
         successfulStx as SmartTransaction,
         {
           chainId: CHAIN_IDS.ETHEREUM,
-          ethersProvider: new Web3Provider(jest.fn()),
+          ethQuery: new EthQuery({ sendAsync: jest.fn() }),
         },
       );
       expect(trackMetaMetricsEventSpy).toHaveBeenCalled();
--- a/src/SmartTransactionsController.test.ts
+++ b/src/SmartTransactionsController.test.ts
@@ -29,7 +29,7 @@ jest.mock('@metamask/eth-query', () => {
     sendAsync = jest.fn(({ method }, callback) => {
       switch (method) {
         case 'getBalance': {
-          callback(null, { toHexString: () => '0x1000' });
+          callback(null, '0x1000');
           break;
         }
 
@@ -38,15 +38,15 @@ jest.mock('@metamask/eth-query', () => {
           break;
         }
 
-        case 'getBlock': {
-          callback(null, { baseFeePerGas: { toHexString: () => '0x123' } });
+        case 'getBlockByNumber': {
+          callback(null, { baseFeePerGas: '0x123' });
           break;
         }
 
         case 'getTransaction': {
           callback(null, {
-            maxFeePerGas: { toHexString: () => '0x123' },
-            maxPriorityFeePerGas: { toHexString: () => '0x123' },
+            maxFeePerGas: '0x123',
+            maxPriorityFeePerGas: '0x123',
           });
           break;
         }
@@ -731,7 +731,9 @@ describe('SmartTransactionsController', () => {
         [CHAIN_IDS.GOERLI]: true,
       });
 
-      await smartTransactionsController.fetchLiveness('goerli');
+      await smartTransactionsController.fetchLiveness({
+        networkClientId: 'goerli',
+      });
 
       expect(
         smartTransactionsController.state.smartTransactionsState
@@ -834,12 +836,6 @@ describe('SmartTransactionsController', () => {
     });
 
     it('throws an error if ethersProvider fails', async () => {
-      // smartTransactionsController.ethQuery.getTransactionReceipt.mockRejectedValueOnce(
-      //   'random error' as never,
-      // );
-      // jest.spyOn('@metmask/eth-query', 'sendAsync')(() => {
-      //   throw new Error('random error');
-      // });
       const successfulStx = {
         ...createStateAfterSuccess()[0],
         history: testHistory,
--- a/src/SmartTransactionsController.test.ts
+++ b/src/SmartTransactionsController.test.ts
@@ -205,7 +205,6 @@ const createStateAfterPending = () => {
       uuid: 'uuid1',
       status: 'pending',
       cancellable: true,
-      chainId: '0x1',
       statusMetadata: {
         cancellationFeeWei: 0,
         cancellationReason: 'not_cancelled',
@@ -234,7 +233,6 @@ const createStateAfterSuccess = () => {
       uuid: 'uuid2',
       status: 'success',
       cancellable: false,
-      chainId: '0x1',
       statusMetadata: {
         cancellationFeeWei: 36777567771000,
         cancellationReason: 'not_cancelled',
@@ -661,6 +659,7 @@ describe('SmartTransactionsController', () => {
       };
       smartTransactionsController.updateSmartTransaction(
         updateTransaction as SmartTransaction,
+        CHAIN_IDS.ETHEREUM,
       );
 
       expect(
@@ -693,6 +692,7 @@ describe('SmartTransactionsController', () => {
       };
       smartTransactionsController.updateSmartTransaction(
         updateTransaction as SmartTransaction,
+        CHAIN_IDS.ETHEREUM,
       );
       expect(confirmSpy).toHaveBeenCalled();
     });
@@ -711,6 +711,7 @@ describe('SmartTransactionsController', () => {
       };
       await smartTransactionsController.confirmSmartTransaction(
         successfulStx as SmartTransaction,
+        CHAIN_IDS.ETHEREUM,
       );
       expect(confirmExternalMock).toHaveBeenCalled();
     });
@@ -724,7 +725,8 @@ describe('SmartTransactionsController', () => {
         history: testHistory,
       };
       await smartTransactionsController.confirmSmartTransaction(
-        successfulStx as SmartTransaction, 
+        successfulStx as SmartTransaction,
+        CHAIN_IDS.ETHEREUM,
       );
       expect(trackMetaMetricsEventSpy).toHaveBeenCalled();
     });
--- a/src/SmartTransactionsController.test.ts
+++ b/src/SmartTransactionsController.test.ts
@@ -258,10 +258,9 @@ const testHistory = [
 ];
 
 const ethereumChainIdDec = parseInt(CHAIN_IDS.ETHEREUM, 16);
+const goerliChainIdDec = parseInt(CHAIN_IDS.GOERLI, 16);
 
 const trackMetaMetricsEventSpy = jest.fn();
-const getNetworkClientByIdSpy = jest.fn();
-
 const defaultState = {
   smartTransactionsState: {
     smartTransactions: {
@@ -307,7 +306,24 @@ describe('SmartTransactionsController', () => {
       provider: jest.fn(),
       confirmExternalTransaction: confirmExternalMock,
       trackMetaMetricsEvent: trackMetaMetricsEventSpy,
-      getNetworkClientById: getNetworkClientByIdSpy,
+      getNetworkClientById: jest.fn().mockImplementation((networkClientId) => {
+        switch (networkClientId) {
+          case 'mainnet':
+            return {
+              configuration: {
+                chainId: CHAIN_IDS.ETHEREUM,
+              },
+            };
+          case 'goerli':
+            return {
+              configuration: {
+                chainId: CHAIN_IDS.GOERLI,
+              },
+            };
+          default:
+            throw new Error('Invalid network client id');
+        }
+      }),
     });
     // eslint-disable-next-line jest/prefer-spy-on
     smartTransactionsController.subscribe = jest.fn();
@@ -402,6 +418,8 @@ describe('SmartTransactionsController', () => {
   });
 
   describe('updateSmartTransactions', () => {
+    // TODO rewrite this test... updateSmartTransactions is getting called via the checkPoll method which is called whenever state is updated.
+    // this test should be more isolated to the updateSmartTransactions method.
     it('calls fetchSmartTransactionsStatus if there are pending transactions', () => {
       const fetchSmartTransactionsStatusSpy = jest
         .spyOn(smartTransactionsController, 'fetchSmartTransactionsStatus')
@@ -490,25 +508,24 @@ describe('SmartTransactionsController', () => {
     });
 
     it('should add fee data to feesByChainId state using the networkClientId passed in to identify the appropriate chain', async () => {
-      const goerliChainIdDec = parseInt(CHAIN_IDS.GOERLI, 16);
-      getNetworkClientByIdSpy.mockImplementation((networkClientId) => {
-        switch (networkClientId) {
-          case 'mainnet':
-            return {
-              configuration: {
-                chainId: CHAIN_IDS.ETHEREUM,
-              },
-            };
-          case 'goerli':
-            return {
-              configuration: {
-                chainId: CHAIN_IDS.GOERLI,
-              },
-            };
-          default:
-            throw new Error('Invalid network client id');
-        }
-      });
+      // getNetworkClientByIdSpy.mockImplementation((networkClientId) => {
+      //   switch (networkClientId) {
+      //     case 'mainnet':
+      //       return {
+      //         configuration: {
+      //           chainId: CHAIN_IDS.ETHEREUM,
+      //         },
+      //       };
+      //     case 'goerli':
+      //       return {
+      //         configuration: {
+      //           chainId: CHAIN_IDS.GOERLI,
+      //         },
+      //       };
+      //     default:
+      //       throw new Error('Invalid network client id');
+      //   }
+      // });
 
       const tradeTx = createUnsignedTransaction(goerliChainIdDec);
       const approvalTx = createUnsignedTransaction(goerliChainIdDec);
@@ -680,6 +697,30 @@ describe('SmartTransactionsController', () => {
         .reply(200, successLivenessApiResponse);
       const liveness = await smartTransactionsController.fetchLiveness();
       expect(liveness).toBe(true);
+    });
+
+    it('fetches liveness and sets in feesByChainId state for the Smart Transactions API for the chainId of the networkClientId passed in', async () => {
+      nock(API_BASE_URL)
+        .get(`/networks/${goerliChainIdDec}/health`)
+        .replyWithError('random error');
+
+      expect(
+        smartTransactionsController.state.smartTransactionsState
+          .livenessByChainId,
+      ).toStrictEqual({
+        [CHAIN_IDS.ETHEREUM]: true,
+        [CHAIN_IDS.GOERLI]: true,
+      });
+
+      await smartTransactionsController.fetchLiveness('goerli');
+
+      expect(
+        smartTransactionsController.state.smartTransactionsState
+          .livenessByChainId,
+      ).toStrictEqual({
+        [CHAIN_IDS.ETHEREUM]: true,
+        [CHAIN_IDS.GOERLI]: false,
+      });
     });
   });
 
@@ -867,7 +908,7 @@ describe('SmartTransactionsController', () => {
   });
 
   describe('startPollingByNetworkClientId', () => {
-    it('starts and stops calling smart transactions batch status api endpoint with the correct chainId at the interval passed via the constructor', async () => {
+    it('starts and stops calling smart transactions batch status api endpoint with the correct chainId at the polling interval', async () => {
       // mock this to a noop because it causes an extra fetch call to the API upon state changes
       jest
         .spyOn(smartTransactionsController, 'checkPoll')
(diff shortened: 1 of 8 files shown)


src/types.ts
```diff
@@ -117,3 +117,5 @@ export type SignedTransaction = any;
 
 // TODO
 export type SignedCanceledTransaction = any;
+
+export type Hex = `0x${string}`;
```
MajorLift, line 121: Could this be imported from `@metamask/utils`?

adonesky1: I suppose we could. I assume this is used somewhere else in the extension dependency graph so probably not adding any weight. I just feel silly adding the full package just to pull in a type here...

adonesky1: cc @mcmire how do you generally think about these cases ☝️ ?

davidmurdoch: Drive by comment: `0x{string}` isn't Hex, and `@metamask/utils` is used by lots of other packages, so this shouldn't be in a general utility library if its is to be named `Hex`. Unbounded (length) hex strings are currently not possible in TypeScript (https://github.com/microsoft/TypeScript/issues/43335).

You can maybe get typescript to full check 30ish characters or so, more if you make use of a "trampoline", but type checking will be so slow it'll be unusable.

src/constants.ts
```diff
@@ -1,6 +1,14 @@
+import { Hex } from './types';
+
 export const API_BASE_URL = 'https://transaction.metaswap.codefi.network';
-export const CHAIN_IDS = {
+export const CHAIN_IDS: {
```
mcmire, line 4: Does it work to let TypeScript infer the type via `as const` instead of manually providing the type?

src/types.ts
mcmire: That's true — the `Hex` type isn't as useful as it could be. For what it's worth we do have [some ways to double check that a string is a real hex string at runtime in `@metamask/utils`](https://github.com/MetaMask/utils/blob/8732f000b7f880cc6d407b23a1f21522dae1b8fc/src/hex.ts), although we might not be using them as much as we should. I tend to think of `Hex` as more of a signal to readers than an actual constraint for consumers.

Since it's basically an approximation and easy to write, I'm okay with copying this and not importing it from `@metamask/utils`. If we need something else from utils then we can always bring the package in later.

src/constants.ts
adonesky1: yes it does 👍

adonesky1: fixed here: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/d644c5ca75e23be8e43c3a59869f7ad381db4a29

src/SmartTransactionsController.ts
```diff
@@ -330,22 +369,27 @@ export default class SmartTransactionsController extends BaseController<
...
-      this.fetchSmartTransactionsStatus(transactionsToUpdate);
+      this.fetchSmartTransactionsStatus(transactionsToUpdate, chainId);
     }
   }
 
-  async confirmSmartTransaction(smartTransaction: SmartTransaction) {
+  // TODO make this a private method?
+  async confirmSmartTransaction(
+    smartTransaction: SmartTransaction,
+    chainId: Hex,
```
mcmire, line 391: This is a breaking change — necessary?

src/SmartTransactionsController.ts
```diff
@@ -420,29 +467,29 @@ export default class SmartTransactionsController extends BaseController<
   // ! Ask backend API to accept list of uuids as params
   async fetchSmartTransactionsStatus(
     uuids: string[],
+    chainId: Hex,
```
mcmire, line 470: This is a breaking change — necessary?

src/SmartTransactionsController.ts
```diff
@@ -252,8 +289,10 @@ export default class SmartTransactionsController extends BaseController<
     return currentIndex === -1 || currentIndex === undefined;
   }
 
-  updateSmartTransaction(smartTransaction: SmartTransaction): void {
-    const { chainId } = this.config;
+  updateSmartTransaction(
+    smartTransaction: SmartTransaction,
+    chainId: Hex,
```
mcmire, line 294: This is a breaking change — necessary?

src/SmartTransactionsController.ts
```diff
@@ -330,22 +369,27 @@ export default class SmartTransactionsController extends BaseController<
     });
   }
 
-  async updateSmartTransactions() {
+  async updateSmartTransactions(
+    networkClientId?: NetworkClientId,
```
mcmire, line 373: Nit: What are your thoughts on placing this in an (optional) options object rather than adding this as a positional argument? I feel like if we ever need to update the signature of this method for any reason then we could do that easily without having to displace this argument. (Same goes for the other cases in this PR where you've added optional arguments.) If this is how you've been doing it in other PRs and are doing it this way for consistency, that's fine, but I just wanted to pose this question in case it was flexible.

adonesky1: That works for me!

adonesky1: > If this is how you've been doing it in other PRs and are doing it this way for consistency, that's fine, but I just wanted to pose this question in case it was flexible.

The pattern hasn't been super consistent across the refactor up to this point. However I think you're correct that an optional options object is the best approach here so I think it's reasonable to adopt that pattern going forward.

src/SmartTransactionsController.ts
adonesky1: same as [here](https://github.com/MetaMask/smart-transactions-controller/pull/210#discussion_r1362749950)

src/SmartTransactionsController.ts
adonesky1: So in this methods case its kindof an annoying situation with this being a public method (it is called directly from the extension) and one used a bunch internally, where a slightly different method signature can be used to avoid a bunch of duplicate calls to `getNetworkByClientId`. So I think my preferred approach here would be to create a wrapper public method that calls a private version of this method and just takes `smartTransactions` and `networkClientId` which we should expect the client to have in the future.

adonesky1: this commit is what I'm thinking for this: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/684bc5923e2e105473626208e2b1ebe71e181928

src/SmartTransactionsController.ts
```diff
@@ -42,23 +44,27 @@ export const DEFAULT_INTERVAL = SECOND * 5;
...
+    smartTransactions: Record<Hex, SmartTransaction[]>;
     userOptIn: boolean | undefined;
     liveness: boolean | undefined;
-    fees: {
-      approvalTxFees: IndividualTxFees | undefined;
-      tradeTxFees: IndividualTxFees | undefined;
-    };
+    fees: FeeEstimates;
+    feesByChainId: Record<Hex, FeeEstimates>;
+    livenessByChainId: Record<Hex, boolean | undefined>;
```
jiexi, line 63:
```suggestion
    livenessByChainId: Record<Hex, boolean>;
```
seems like we can always ensure that we set a bool?

src/SmartTransactionsController.ts
```diff
@@ -520,6 +608,22 @@ export default class SmartTransactionsController extends BaseController<
         },
       },
     });
+
+    if (networkClientId) {
```
jiexi, line 612: does it make a difference if we populated feesByChainId even if networkClientId isn't provided?

src/SmartTransactionsController.ts
```diff
@@ -566,36 +671,55 @@ export default class SmartTransactionsController extends BaseController<
...
+  }
+
   // TODO: This should return if the cancellation was on chain or not (for nonce management)
   // After this successful call client must update nonce representative
   // in transaction controller external transactions list
-  async cancelSmartTransaction(uuid: string): Promise<void> {
-    const { chainId } = this.config;
+  async cancelSmartTransaction(
+    uuid: string,
+    networkClientId?: NetworkClientId,
```
jiexi, line 712: this interface is odd. I would expect chainId to come off the transaction object itself. Not asking for any change, just pointing this out

src/SmartTransactionsController.ts
adonesky1: Yeah I suppose not. Presumably we won't start accessing that state until networkClientId is the passed everywhere and the globally selected network version of the state is deprecated. But yeah we can add it to the `feesByChainId` regardless for now.

src/SmartTransactionsController.ts
adonesky1: In this case there is no transaction object? Just a uuid

src/SmartTransactionsController.ts
```diff
@@ -330,32 +396,48 @@ export default class SmartTransactionsController extends BaseController<
...
-  async confirmSmartTransaction(smartTransaction: SmartTransaction) {
+  // TODO make this a private method?
+  async confirmSmartTransaction(
+    smartTransaction: SmartTransaction,
+    {
+      chainId,
+      ethersProvider,
+    }: {
+      chainId: Hex;
+      ethersProvider: any;
```
mcmire, line 427: The type here should be this (where you can get `Provider` from NetworkController):

``` suggestion
      ethersProvider: Web3Provider | Provider;

File: src/SmartTransactionsController.ts [hunk 11]
```diff
@@ -330,32 +396,48 @@ export default class SmartTransactionsController extends BaseController<
...
+      const transactionReceipt = await ethersProvider.getTransactionReceipt(
+        txHash,
+      );
+      const transaction = await ethersProvider.getTransaction(txHash);
       const maxFeePerGas = transaction.maxFeePerGas?.toHexString();
       const maxPriorityFeePerGas =
         transaction.maxPriorityFeePerGas?.toHexString();
       if (transactionReceipt?.blockNumber) {
-        const blockData = await this.ethersProvider.getBlock(
+        const blockData = await ethersProvider.getBlock(
```
mcmire, line 440: Hmm... this isn't going to work anymore. We can't guarantee that `ethersProvider` is an Ethers provider; it could be our provider from NetworkController, in which case methods like this aren't supported.

Do we even have to use Web3Provider in this file? We have typically use `@metamask/eth-query`, which just uses `sendAsync` to make requests.

mcmire (changes requested):
It seems like there are now two different providers that this controller could use. This is currently being hidden by the use of `any` in this file, but once you lift that you should see. We'll have to account for this, or switch from Ethers to `@metamask/eth-query`.

src/SmartTransactionsController.ts
mcmire: I'll do another review round later so I can see it in context, but your change makes sense I believe!

src/SmartTransactionsController.ts
adonesky1: Yeah I could sub out `Web3Provider` for `@metamask/eth-query` throughout this controller if that is the direction we've been heading?

src/SmartTransactionsController.ts
```diff
@@ -420,29 +505,31 @@ export default class SmartTransactionsController extends BaseController<
...
+    { networkClientId }: { networkClientId?: NetworkClientId } = {},
   ): Promise<SmartTransaction[]> {
-    const { chainId } = this.config;
-
     const params = new URLSearchParams({
       uuids: uuids.join(','),
     });
-
+    const chainId = this.getChainId(networkClientId);
+    const ethersProvider = this.getProvider(networkClientId);
```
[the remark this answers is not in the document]
adonesky1: swapped out ethersprovider altogether: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/fe9732ccf000b24a670fca8681edc998f45771c2

src/SmartTransactionsController.ts
adonesky1: swapped out ethersprovider altogether: https://github.com/MetaMask/smart-transactions-controller/commit/fe9732ccf000b24a670fca8681edc998f45771c2

src/SmartTransactionsController.ts
```diff
@@ -252,8 +289,34 @@ export default class SmartTransactionsController extends BaseController<
...
+  }
+
+  #updateSmartTransaction(
+    smartTransaction: SmartTransaction,
+    {
+      chainId = this.config.chainId,
+      ethersProvider = this.ethersProvider,
+    }: {
+      chainId: Hex;
+      ethersProvider: Web3Provider;
```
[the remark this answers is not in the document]
adonesky1: swapped out ethersprovider altogether: https://github.com/MetaMask/smart-transactions-controller/commit/fe9732ccf000b24a670fca8681edc998f45771c2

src/SmartTransactionsController.ts
adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/commit/fe9732ccf000b24a670fca8681edc998f45771c2

src/SmartTransactionsController.ts
```diff
@@ -566,36 +663,59 @@ export default class SmartTransactionsController extends BaseController<
...
+    }
+    return this.getNetworkClientById(networkClientId).configuration.chainId;
+  }
+
+  getProvider(networkClientId?: NetworkClientId): any {
+    if (!networkClientId) {
+      return this.ethersProvider;
+    }
+    return this.getNetworkClientById(networkClientId).provider;
+  }
```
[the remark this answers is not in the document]
adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/fe9732ccf000b24a670fca8681edc998f45771c2

src/SmartTransactionsController.ts
```diff
@@ -566,36 +663,59 @@ export default class SmartTransactionsController extends BaseController<
...
 
     return data;
   }
 
+  getChainId(networkClientId?: NetworkClientId): Hex {
+    if (!networkClientId) {
+      return this.config.chainId;
+    }
+    return this.getNetworkClientById(networkClientId).configuration.chainId;
+  }
```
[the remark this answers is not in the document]
adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/fe9732ccf000b24a670fca8681edc998f45771c2

src/SmartTransactionsController.test.ts
```diff
@@ -633,20 +825,31 @@ describe('SmartTransactionsController', () => {
...
+        },
       );
       expect(confirmExternalMock).toHaveBeenCalled();
     });
 
     it('throws an error if ethersProvider fails', async () => {
-      smartTransactionsController.ethersProvider.getTransactionReceipt.mockRejectedValueOnce(
-        'random error' as never,
-      );
+      // smartTransactionsController.ethQuery.getTransactionReceipt.mockRejectedValueOnce(
```
mcmire, line 837: ✂️ ?

src/SmartTransactionsController.ts
```diff
@@ -330,35 +397,54 @@ export default class SmartTransactionsController extends BaseController<
...
+        [txHash],
+      );
+      const transaction = await query(ethQuery, 'getTransaction', [txHash]);
+
       const maxFeePerGas = transaction.maxFeePerGas?.toHexString();
       const maxPriorityFeePerGas =
         transaction.maxPriorityFeePerGas?.toHexString();
       if (transactionReceipt?.blockNumber) {
-        const blockData = await this.ethersProvider.getBlock(
+        const blockData = await query(ethQuery, 'getBlock', [
```
mcmire, line 444: It seems that Ethers has a method called `getBlock` that either calls `eth_getBlockByNumber` or `eth_getBlockByHash` depending on the argument: https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/json-rpc-provider.ts#L578

Also, `query` returns `any` (sigh).

To fix this and make this more type-safe, maybe we say:

``` suggestion
        const blockData: { baseFeePerGas?: string } | null = await query(ethQuery, 'getBlockByNumber', [
```

src/SmartTransactionsController.ts
```diff
@@ -330,35 +397,54 @@ export default class SmartTransactionsController extends BaseController<
...
           transactionReceipt?.blockNumber,
           false,
-        );
+        ]);
         const baseFeePerGas = blockData?.baseFeePerGas.toHexString();
```
mcmire, line 448: This also won't work anymore since this is just a string (also `baseFeePerGas` may not be there if the network doesn't support EIP-1559). However, this is already a hex string anyway so we can just take it off:

``` suggestion
        const baseFeePerGas = blockData?.baseFeePerGas;
```

src/SmartTransactionsController.ts
```diff
@@ -330,35 +397,54 @@ export default class SmartTransactionsController extends BaseController<
...
         const txReceipt = mapValues(transactionReceipt, (value) => {
```
mcmire, line 449: We should be able to simplify this to `Object.values(transactionReceipt)` (`value` will never be an instanceof ethersBigNumber).

src/SmartTransactionsController.ts
```diff
@@ -330,35 +397,54 @@ export default class SmartTransactionsController extends BaseController<
...
+      chainId: Hex;
+      ethQuery: EthQuery;
+    },
+  ) {
     const txHash = smartTransaction.statusMetadata?.minedHash;
     try {
-      const transactionReceipt =
-        await this.ethersProvider.getTransactionReceipt(txHash);
-      const transaction = await this.ethersProvider.getTransaction(txHash);
+      const transactionReceipt = await query(
```
mcmire, line 433: `query` returns `any`, so we should give it a type:

``` suggestion
      const transactionReceipt: { maxFeePerGas?: string; maxPriorityFeePerGas?: string; blockNumber: string } | null = await query(

File: src/SmartTransactionsController.ts [hunk 21]
```diff
@@ -566,36 +670,64 @@ export default class SmartTransactionsController extends BaseController<
...
+    const chainId = this.getChainId({ networkClientId });
     await this.fetch(getAPIRequestURL(APIType.CANCEL, chainId), {
       method: 'POST',
       body: JSON.stringify({ uuid }),
     });
   }
 
-  async fetchLiveness(): Promise<boolean> {
-    const { chainId } = this.config;
+  async fetchLiveness(networkClientId?: NetworkClientId): Promise<boolean> {
```
mcmire, line 729: Do we want to put this in an options object too?

src/SmartTransactionsController.ts
```diff
@@ -252,8 +289,35 @@ export default class SmartTransactionsController extends BaseController<
...
+  updateSmartTransaction(
+    smartTransaction: SmartTransaction,
+    { networkClientId }: { networkClientId?: NetworkClientId } = {},
+  ) {
+    let { chainId } = this.config;
+    let { ethQuery } = this;
+    if (networkClientId) {
+      const networkClient = this.getNetworkClientById(networkClientId);
+      chainId = networkClient.configuration.chainId;
+      // @ts-expect-error TODO: Provider type alignment
```
mcmire, line 301: Oh yeah :( Well, we'll fix this soon hopefully

mcmire (commented):
Seems like there's some leftover Ethers code still that may cause bugs.

src/SmartTransactionsController.ts
adonesky1: Done here: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/94e2fac1cfdfedacd31b038a72815abb013dc19c

src/SmartTransactionsController.ts
adonesky1: Done here: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/94e2fac1cfdfedacd31b038a72815abb013dc19c

src/SmartTransactionsController.ts
adonesky1: Done here: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/94e2fac1cfdfedacd31b038a72815abb013dc19c

src/SmartTransactionsController.ts
adonesky1: Done here: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/94e2fac1cfdfedacd31b038a72815abb013dc19c

src/SmartTransactionsController.ts
adonesky1: Done here: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/94e2fac1cfdfedacd31b038a72815abb013dc19c

src/SmartTransactionsController.test.ts
adonesky1: Done here: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/94e2fac1cfdfedacd31b038a72815abb013dc19c

adonesky1:
Time to update jest https://github.com/MetaMask/smart-transactions-controller/pull/210#issuecomment-1769150910 @mcmire ?

adonesky1:
@SocketSecurity ignore ws@7.5.9

src/SmartTransactionsController.ts
```diff
@@ -330,42 +395,60 @@ export default class SmartTransactionsController extends BaseController<
...
     if (transactionsToUpdate.length > 0) {
-      this.fetchSmartTransactionsStatus(transactionsToUpdate);
+      this.fetchSmartTransactionsStatus(transactionsToUpdate, {
+        networkClientId,
+      });
     }
   }
 
-  async confirmSmartTransaction(smartTransaction: SmartTransaction) {
+  // TODO make this a private method?
```
jiexi, line 418: 🤔

jiexi (dismissed):

src/SmartTransactionsController.ts
mcmire: I think we should double-check that this isn't used anywhere. Regardless it seems okay to leave for now. Maybe we remove this TODO for now (unless we are getting to it soon) and ticket this instead? Your call on how you want to keep track of it though

src/SmartTransactionsController.ts
```diff
@@ -477,8 +566,9 @@ export default class SmartTransactionsController extends BaseController<
   async getFees(
     tradeTx: UnsignedTransaction,
     approvalTx: UnsignedTransaction,
+    networkClientId?: NetworkClientId,
```
mcmire, line 569: Do we want to put this in an options object too?

src/SmartTransactionsController.ts
mcmire: Update: I just double-checked this myself. Let's make this private!

mcmire (commented):
Getting closer!

src/SmartTransactionsController.ts
adonesky1: how do I keep missing these 😮‍💨

mcmire: There's a lot in this one!

adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/99fc2cc69f1b6a520d23b3eb77f092c883305a78

src/SmartTransactionsController.ts
```diff
@@ -330,42 +396,59 @@ export default class SmartTransactionsController extends BaseController<
...
+      const transactionReceipt: {
+        maxFeePerGas?: string;
+        maxPriorityFeePerGas?: string;
+        blockNumber: string;
+      } | null = await query(ethQuery, 'getTransactionReceipt', [txHash]);
+
+      const transaction: {
+        maxFeePerGas?: string;
+        maxPriorityFeePerGas?: string;
+      } = await query(ethQuery, 'getTransaction', [txHash]);
```
mcmire, line 440: Looks like `getTransaction` was another Ethers convenience method. Should this be `getTransactionByHash`? If so then Infura says [that this can also be `null`](https://docs.infura.io/networks/ethereum/json-rpc-methods/eth_gettransactionbyhash):

``` suggestion
      } | null = await query(ethQuery, 'getTransactionByHash', [txHash]);

File: src/SmartTransactionsController.ts (hunk 23 above)
Comment (adonesky1) [reply src/SmartTransactionsController.ts]: done here: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/99fc2cc69f1b6a520d23b3eb77f092c883305a78

File: src/SmartTransactionsController.test.ts [hunk 26]
```diff
@@ -1,44 +1,74 @@
...
-    }));
-
-    getBlock = jest.fn();
-  },
-}));
+jest.mock('@metamask/eth-query', () => {
+  return class FakeEthQuery {
+    sendAsync = jest.fn(({ method }, callback) => {
+      switch (method) {
+        case 'getBalance': {
```
mcmire, line 30: Should we prepend these methods with `eth_`? Also, I believe `getTransaction` should be `eth_getTransactionByHash` (see comment I left below).

mcmire (commented):
One more thing and then I think we're good 😅

src/SmartTransactionsController.test.ts
adonesky1: Looks like there are a bunch of instances in various controllers where the methods are not [not prepended with `eth_`](https://github.com/MetaMask/core/blob/d992b424f934c4a96983c9a5005a44958e02979f/packages/assets-controllers/src/AccountTrackerController.ts#L197). Other examples [here](https://github.com/MetaMask/core/blob/d992b424f934c4a96983c9a5005a44958e02979f/packages/gas-fee-controller/src/gas-util.ts#L123), [here](https://github.com/MetaMask/core/blob/d992b424f934c4a96983c9a5005a44958e02979f/packages/gas-fee-controller/src/fetchGasEstimatesViaEthFeeHistory/fetchLatestBlock.ts#L18) & [here](https://github.com/MetaMask/core/blob/d992b424f934c4a96983c9a5005a44958e02979f/packages/transaction-controller/src/utils/gas.ts#L69)

adonesky1: Also seems like you could just call the methods right off the ethQuery class if you wanted based [on the readme](https://github.com/MetaMask/eth-query#readme)

src/SmartTransactionsController.ts
adonesky1: 👍 fixed here: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/9c4a772cc5dbff42e11bdad85d1076b94f751c9d

src/SmartTransactionsController.test.ts
mcmire: Ah, different definitions of "method" :)

The _instance_ methods on EthQuery aren't prepended with `eth_`, but when the request is made through `sendAsync`, the `eth_` prefix is added to the _RPC_ method. So if we are mocking `sendAsync` on EthQuery, it seems like we ought to use the correct RPC method. Otherwise, I'm not sure how these tests are passing 🤔

mcmire: Okay, I dug into this a bit. When using `query` in the implementation file, we're passing the name of an EthQuery instance method, and so `query` will then try to call a method on the EthQuery instance that matches that name. But because the only method we have in our fake EthQuery class is `sendAsync`, `query` will then call `sendAsync` and just pass along the method it has: https://github.com/MetaMask/core/blob/306eecda2ef1d0ebcf4df3b8e56338c0f16ceac9/packages/controller-utils/src/util.ts#L432-L454. Hence, `method` here _doesn't_ have an `eth_` prefix.

This accidentally works because none of the strings in the `switch` statement have that prefix. But I think it might be confusing if people are familiar with how EthQuery is supposed to work.

Therefore would it make more sense to bring in `EthQuery` via `jest.requireActual`, `extend` from it, and then update all of these RPC methods to have an `eth_` prefix? (I can make a new suggestion)

src/SmartTransactionsController.test.ts
```diff
@@ -1,44 +1,74 @@
...
+          callback(null, { blockNumber: '123' });
+          break;
+        }
+
+        case 'getBlockByNumber': {
+          callback(null, { baseFeePerGas: '0x123' });
+          break;
+        }
+
+        case 'getTransactionByHash': {
```
mcmire, line 45: As per question above:

``` suggestion
jest.mock('@metamask/eth-query', () => {
  const EthQuery = jest.requireActual('@metamask/eth-query');

  return class FakeEthQuery extends EthQuery {
    sendAsync = jest.fn(({ method }, callback) => {
      switch (method) {
        case 'eth_getBalance': {
          callback(null, '0x1000');
          break;
        }

        case 'eth_getTransactionReceipt': {
          callback(null, { blockNumber: '123' });
          break;
        }

        case 'eth_getBlockByNumber': {
          callback(null, { baseFeePerGas: '0x123' });
          break;
        }

        case 'eth_getTransactionByHash': {
```

adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/210/commits/c74619c6f4c284dd07b14b00a690ea8e65aca922

mcmire (approved):
Thanks for your patience — everything looks good now! (I was getting the test failure locally — I don't think requests are being mocked in this package fully. Should be able to re-run.)

PR #222, opened 2023-11-03
Add a new "userOptInV2" prop

dan437:
## Description
We are adding a new "userOptInV2" prop, since we are updating our backend and need users to re-approve Smart Swaps.


infiniteflower (approved):
LGTM

PR #227, opened 2023-11-07
Stx controller version

dan437:
## Description
Add the "stxControllerVersion" query param for the `/submitTransactions` API call. That's because then backend can have custom logic based on the version.


cloudonshore (approved):

infiniteflower (approved):
LGTM

PR #237, opened 2023-11-09
Parameterize SmartTransactionsController state by ChainId for MultiChain + Integrate PollingController Mixin

adonesky1:
~TODO:~
- [x] ~E2E test with extension to ensure working properly before taking out of draft~

Working e2e test here:
https://github.com/MetaMask/smart-transactions-controller/assets/34557516/307de2da-1fe3-43c8-8faa-7bf51da28eaa

Resolves: https://github.com/MetaMask/MetaMask-planning/issues/1039

Integrates new `PollingController` abstraction into the `SmartTransactionsController`, implements the `_executePoll` method and parameterizes the `updateSmartTransactions` method by `networkClientId` such that pending SmartTransaction statuses can be polled for concurrently across different chains.

A note: As is our current controller refactor strategy, these changes are additive, backwards compatible and coexist alongside the existing globally selected network pattern. Once we fully adopt this polling pattern we should no longer access the root liveness and fees state but rather access these values by chainId from `livenessByChainId` and `feesByChainId`.

### Added
- Integrates `PollingController` mixin into the `SmartTransactionsController` and implements the abstract `_executePoll` method.
### Changed
- Adds optional options object containing a `networkClientId` argument to the `updateSmartTransaction` method which, if passed, is used fetch the chainId that corresponds to the networkClientId and is then used to fetch and update the status of pending smart transactions for that chainId.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,12 +25,12 @@
     "test:watch": "jest --watchAll"
   },
   "dependencies": {
-    "@ethersproject/bignumber": "^5.7.0",
     "@ethersproject/bytes": "^5.7.0",
-    "@ethersproject/providers": "^5.7.0",
     "@metamask/base-controller": "^3.2.1",
     "@metamask/controller-utils": "^6.0.0",
     "@metamask/network-controller": "^17.0.0",
+    "@metamask/eth-query": "^3.0.1",
+    "@metamask/polling-controller": "^1.0.1",
     "bignumber.js": "^9.0.1",
     "fast-json-patch": "^3.1.0",
     "lodash": "^4.17.21"
@@ -45,6 +45,7 @@
     "@types/jest": "^26.0.24",
     "@types/lodash": "^4.14.194",
     "@types/node": "^16.18.31",
+    "@types/sinon": "^9.0.10",
     "@typescript-eslint/eslint-plugin": "^5.33.0",
     "@typescript-eslint/parser": "^5.33.0",
     "eslint": "^8.21.0",
@@ -59,6 +60,7 @@
     "nock": "^13.3.1",
     "prettier": "^2.8.8",
     "prettier-plugin-packagejson": "^2.4.3",
+    "sinon": "^9.2.4",
     "ts-jest": "^26.5.6",
     "typescript": "~4.4.4"
   },
--- a/package.json
+++ b/package.json
@@ -26,11 +26,11 @@
   },
   "dependencies": {
     "@ethersproject/bytes": "^5.7.0",
-    "@metamask/network-controller": "^17.0.0",
     "@metamask/base-controller": "^4.0.0",
     "@metamask/controller-utils": "^6.1.0",
     "@metamask/eth-query": "^4.0.0",
-    "@metamask/polling-controller": "^1.0.1",
+    "@metamask/network-controller": "^17.0.0",
+    "@metamask/polling-controller": "^2.0.0",
     "bignumber.js": "^9.0.1",
     "fast-json-patch": "^3.1.0",
     "lodash": "^4.17.21"
--- a/package.json
+++ b/package.json
@@ -26,10 +26,10 @@
   },
   "dependencies": {
     "@ethersproject/bytes": "^5.7.0",
-    "@metamask/base-controller": "^3.2.1",
-    "@metamask/controller-utils": "^6.0.0",
     "@metamask/network-controller": "^17.0.0",
-    "@metamask/eth-query": "^3.0.1",
+    "@metamask/base-controller": "^4.0.0",
+    "@metamask/controller-utils": "^6.1.0",
+    "@metamask/eth-query": "^4.0.0",
     "@metamask/polling-controller": "^1.0.1",
     "bignumber.js": "^9.0.1",
     "fast-json-patch": "^3.1.0",
--- a/src/SmartTransactionsController.test.ts
+++ b/src/SmartTransactionsController.test.ts
@@ -938,57 +938,50 @@ describe('SmartTransactionsController', () => {
 
       await advanceTime({ clock, duration: 0 });
 
-      expect(
-        JSON.stringify(handleFetchSpy.mock.calls.map((arg) => arg[0])),
-      ).toStrictEqual(
-        JSON.stringify([
-          `${API_BASE_URL}/networks/${convertHexToDecimal(
-            CHAIN_IDS.ETHEREUM,
-          )}/batchStatus?uuids=uuid1`,
-        ]),
+      const fetchHeaders = {
+        headers: {
+          'Content-Type': 'application/json',
+          'X-Client-Id': 'default',
+        },
+      };
+
+      expect(handleFetchSpy).toHaveBeenNthCalledWith(
+        1,
+        `${API_BASE_URL}/networks/${convertHexToDecimal(
+          CHAIN_IDS.ETHEREUM,
+        )}/batchStatus?uuids=uuid1`,
+        fetchHeaders,
       );
 
-      handleFetchSpy.mockClear();
       await advanceTime({ clock, duration: DEFAULT_INTERVAL });
 
-      expect(
-        JSON.stringify(handleFetchSpy.mock.calls.map((arg) => arg[0])),
-      ).toStrictEqual(
-        JSON.stringify([
-          `${API_BASE_URL}/networks/${convertHexToDecimal(
-            CHAIN_IDS.ETHEREUM,
-          )}/batchStatus?uuids=uuid1`,
-        ]),
+      expect(handleFetchSpy).toHaveBeenNthCalledWith(
+        2,
+        `${API_BASE_URL}/networks/${convertHexToDecimal(
+          CHAIN_IDS.ETHEREUM,
+        )}/batchStatus?uuids=uuid1`,
+        fetchHeaders,
       );
 
-      handleFetchSpy.mockClear();
       smartTransactionsController.startPollingByNetworkClientId('goerli');
       await advanceTime({ clock, duration: 0 });
 
-      expect(
-        JSON.stringify(handleFetchSpy.mock.calls.map((arg) => arg[0])),
-      ).toStrictEqual(
-        JSON.stringify([
-          `${API_BASE_URL}/networks/${convertHexToDecimal(
-            CHAIN_IDS.GOERLI,
-          )}/batchStatus?uuids=uuid2`,
-        ]),
+      expect(handleFetchSpy).toHaveBeenNthCalledWith(
+        3,
+        `${API_BASE_URL}/networks/${convertHexToDecimal(
+          CHAIN_IDS.GOERLI,
+        )}/batchStatus?uuids=uuid2`,
+        fetchHeaders,
       );
 
-      handleFetchSpy.mockClear();
       await advanceTime({ clock, duration: DEFAULT_INTERVAL });
 
-      expect(
-        JSON.stringify(handleFetchSpy.mock.calls.map((arg) => arg[0])),
-      ).toStrictEqual(
-        JSON.stringify([
-          `${API_BASE_URL}/networks/${convertHexToDecimal(
-            CHAIN_IDS.ETHEREUM,
-          )}/batchStatus?uuids=uuid1`,
-          `${API_BASE_URL}/networks/${convertHexToDecimal(
-            CHAIN_IDS.GOERLI,
-          )}/batchStatus?uuids=uuid2`,
-        ]),
+      expect(handleFetchSpy).toHaveBeenNthCalledWith(
+        5,
+        `${API_BASE_URL}/networks/${convertHexToDecimal(
+          CHAIN_IDS.GOERLI,
+        )}/batchStatus?uuids=uuid2`,
+        fetchHeaders,
       );
 
       // stop the mainnet polling
@@ -997,23 +990,25 @@ describe('SmartTransactionsController', () => {
       );
 
       // cycle two polling intervals
-      handleFetchSpy.mockClear();
       await advanceTime({ clock, duration: DEFAULT_INTERVAL });
 
       await advanceTime({ clock, duration: DEFAULT_INTERVAL });
 
       // check that the mainnet polling has stopped while the goerli polling continues
-      expect(
-        JSON.stringify(handleFetchSpy.mock.calls.map((arg) => arg[0])),
-      ).toStrictEqual(
-        JSON.stringify([
-          `${API_BASE_URL}/networks/${convertHexToDecimal(
-            CHAIN_IDS.GOERLI,
-          )}/batchStatus?uuids=uuid2`,
-          `${API_BASE_URL}/networks/${convertHexToDecimal(
-            CHAIN_IDS.GOERLI,
-          )}/batchStatus?uuids=uuid2`,
-        ]),
+      expect(handleFetchSpy).toHaveBeenNthCalledWith(
+        6,
+        `${API_BASE_URL}/networks/${convertHexToDecimal(
+          CHAIN_IDS.GOERLI,
+        )}/batchStatus?uuids=uuid2`,
+        fetchHeaders,
+      );
+
+      expect(handleFetchSpy).toHaveBeenNthCalledWith(
+        7,
+        `${API_BASE_URL}/networks/${convertHexToDecimal(
+          CHAIN_IDS.GOERLI,
+        )}/batchStatus?uuids=uuid2`,
+        fetchHeaders,
       );
 
       // cleanup
--- a/src/SmartTransactionsController.test.ts
+++ b/src/SmartTransactionsController.test.ts
@@ -323,7 +323,7 @@ describe('SmartTransactionsController', () => {
           releaseLock: jest.fn(),
         };
       }),
-      provider: jest.fn(),
+      provider: { sendAsync: jest.fn() },
       confirmExternalTransaction: jest.fn(),
       trackMetaMetricsEvent: trackMetaMetricsEventSpy,
       getNetworkClientById: jest.fn().mockImplementation((networkClientId) => {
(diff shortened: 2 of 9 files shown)


src/SmartTransactionsController.ts
```diff
@@ -423,29 +512,31 @@ export default class SmartTransactionsController extends BaseController<
...
     });
-
+    const chainId = this.getChainId({ networkClientId });
+    const ethQuery = this.getEthQuery({ networkClientId });
     const url = `${getAPIRequestURL(
       APIType.BATCH_STATUS,
       chainId,
     )}?${params.toString()}`;
 
     const data = await this.fetch(url);
```
MajorLift, line 527:
```suggestion
    const data = (await this.fetch(url)) as Record<
      string,
      SmartTransactionsStatus
    >;
```

src/SmartTransactionsController.ts
```diff
@@ -423,29 +512,31 @@ export default class SmartTransactionsController extends BaseController<
...
-        ),
-        uuid,
-      });
+      this.#updateSmartTransaction(
+        {
+          statusMetadata: stxStatus as SmartTransactionsStatus,
+          status: calculateStatus(stxStatus as SmartTransactionsStatus),
+          cancellable: isSmartTransactionCancellable(
+            stxStatus as SmartTransactionsStatus,
+          ),
```
MajorLift, line 535:
```suggestion
          statusMetadata: stxStatus,
          status: calculateStatus(stxStatus),
          cancellable: isSmartTransactionCancellable(stxStatus),
```

MajorLift: Follows from https://github.com/MetaMask/smart-transactions-controller/pull/237/files#r1402633206, which lets `stxStatus` be inferred.

src/SmartTransactionsController.ts
```diff
@@ -423,29 +512,31 @@ export default class SmartTransactionsController extends BaseController<
   // ! Ask backend API to accept list of uuids as params
   async fetchSmartTransactionsStatus(
     uuids: string[],
+    { networkClientId }: { networkClientId?: NetworkClientId } = {},
   ): Promise<SmartTransaction[]> {
```
MajorLift, line 516: This return type seems to be inaccurate.

```suggestion
  ): Promise<Record<string, SmartTransactionsStatus>> {
```

src/SmartTransactionsController.ts
```diff
@@ -333,42 +398,58 @@ export default class SmartTransactionsController extends BaseController<
...
-  async updateSmartTransactions() {
+  async updateSmartTransactions({
+    networkClientId,
+  }: {
+    networkClientId?: NetworkClientId;
+  } = {}): Promise<void> {
     const { smartTransactions } = this.state.smartTransactionsState;
-    const { chainId } = this.config;
+    const chainId = this.getChainId({ networkClientId });
+    const smartTransactionsForChainId = smartTransactions?.[chainId];
```
MajorLift, line 408:
```suggestion
    const smartTransactionsForChainId = smartTransactions[chainId];
```

This can be removed without causing errors.

src/SmartTransactionsController.ts
```diff
@@ -103,6 +109,7 @@ export default class SmartTransactionsController extends BaseController<
       provider: any;
```
MajorLift, line 109:
```suggestion
      provider: Provider;
```

This can be done without disrupting anything, since the errors caused by this are already suppressed by `// @ts-expect-error TODO: Provider type alignment` directives.

src/test-helpers.ts
```diff
@@ -0,0 +1,37 @@
...
+ * this function ensures that both timers and promises are comprehensively processed.
+ * @param options - The options object.
+ * @param options.clock - The Sinon fake timer instance used to manipulate time in tests.
+ * @param options.duration - The total amount of time (in milliseconds) to advance the timer by.
+ * @param options.stepSize - The incremental step size (in milliseconds) by which the timer is advanced in each iteration. Default is 1/4 of the duration.
+ */
+export async function advanceTime({
+  clock,
+  duration,
+  stepSize = duration / 4,
```
MajorLift, line 25: Nit: just in case `duration` is passed into something that expects an integer.
```suggestion
  stepSize = Math.round(duration / 4),
```

package.json
```diff
@@ -25,12 +25,12 @@
     "test:watch": "jest --watchAll"
   },
   "dependencies": {
-    "@ethersproject/bignumber": "^5.7.0",
     "@ethersproject/bytes": "^5.7.0",
-    "@ethersproject/providers": "^5.7.0",
     "@metamask/base-controller": "^3.2.1",
     "@metamask/controller-utils": "^5.0.0",
+    "@metamask/eth-query": "^3.0.1",
```
MajorLift, line 31:
```suggestion
    "@metamask/base-controller": "^4.0.0",
    "@metamask/controller-utils": "^6.1.0",
    "@metamask/eth-query": "^4.0.0",
```

With these version bumps (and https://github.com/MetaMask/smart-transactions-controller/pull/237#discussion_r1402649548), all `// @ts-expect error TODO: Provider type alignment` directives can be removed.

src/test-helpers.ts
adonesky1: It will only ever be passed to `clock.tickAsync` but I suppose this is fine

adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/237/commits/93a05f14cbf91a344332dcf07d39f6924d5eacd3

src/SmartTransactionsController.ts
adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/237/commits/93a05f14cbf91a344332dcf07d39f6924d5eacd3

src/SmartTransactionsController.ts
adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/237/commits/93a05f14cbf91a344332dcf07d39f6924d5eacd3

src/SmartTransactionsController.ts
adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/237/commits/93a05f14cbf91a344332dcf07d39f6924d5eacd3

src/SmartTransactionsController.ts
adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/237/commits/93a05f14cbf91a344332dcf07d39f6924d5eacd3

src/SmartTransactionsController.test.ts
```diff
@@ -9,40 +11,63 @@ import SmartTransactionsController, {
 } from './SmartTransactionsController';
 import { API_BASE_URL, CHAIN_IDS } from './constants';
 import { SmartTransaction, SmartTransactionStatuses } from './types';
+import * as utils from './utils';
+import { flushPromises, advanceTime } from './test-helpers';
 
-const confirmExternalMock = jest.fn();
+// const confirmExternalMock = jest.fn();
```
jiexi, line 17:
```suggestion
```

src/SmartTransactionsController.ts
```diff
@@ -255,8 +289,34 @@ export default class SmartTransactionsController extends BaseController<
...
   }
 
-  updateSmartTransaction(smartTransaction: SmartTransaction): void {
-    const { chainId } = this.config;
+  updateSmartTransaction(
+    smartTransaction: SmartTransaction,
+    { networkClientId }: { networkClientId?: NetworkClientId } = {},
+  ) {
+    let { chainId } = this.config;
+    let { ethQuery } = this;
```
jiexi, line 297:
```suggestion
    let { ethQuery, config: { chainId } = this;
```

ehh idk. `let { ethQuery } = this;` specifically feels funky, but i guess that's only because you aren't usually deconstructing `this`

jiexi: maybe even `let chainId, ethQuery;` since you default it in `#updateSmartTransaction` already

src/SmartTransactionsController.ts
```diff
@@ -289,7 +350,7 @@ export default class SmartTransactionsController extends BaseController<
               .slice(0, cancelledNonceIndex)
               .concat(currentSmartTransactions.slice(cancelledNonceIndex + 1))
               .concat(historifiedSmartTransaction)
-          : currentSmartTransactions.concat(historifiedSmartTransaction);
+          : currentSmartTransactions?.concat(historifiedSmartTransaction);
```
jiexi, line 353: `nextSmartTransactions` would be undefined in state rather than an empty array if  `currentSmartTransactions` is undefined. Is that what we want?

src/SmartTransactionsController.ts
```diff
@@ -255,8 +289,34 @@ export default class SmartTransactionsController extends BaseController<
...
+    smartTransaction: SmartTransaction,
+    { networkClientId }: { networkClientId?: NetworkClientId } = {},
+  ) {
+    let { chainId } = this.config;
+    let { ethQuery } = this;
+    if (networkClientId) {
+      const networkClient = this.getNetworkClientById(networkClientId);
+      chainId = networkClient.configuration.chainId;
+      ethQuery = new EthQuery(networkClient.provider);
+    }
```
jiexi, line 302: are these helpers useful here?

```
    const chainId = this.getChainId({ networkClientId });
    const ethQuery = this.getEthQuery({ networkClientId });
```

src/SmartTransactionsController.ts
```diff
@@ -521,8 +614,16 @@ export default class SmartTransactionsController extends BaseController<
           approvalTxFees,
           tradeTxFees,
         },
```
jiexi, line 616: you only want to set `fees` if it matches the networkClientId matches the current globally selected chainId, right?

src/SmartTransactionsController.ts
```diff
@@ -569,36 +673,68 @@ export default class SmartTransactionsController extends BaseController<
...
+    return networkClientId
+      ? this.getNetworkClientById(networkClientId).configuration.chainId
+      : this.config.chainId;
+  }
+
+  getEthQuery({
+    networkClientId,
+  }: {
+    networkClientId?: NetworkClientId;
+  }): EthQuery {
```
jiexi, line 708: does this need a default value `= {}`

src/SmartTransactionsController.ts
```diff
@@ -613,8 +749,13 @@ export default class SmartTransactionsController extends BaseController<
       smartTransactionsState: {
         ...this.state.smartTransactionsState,
         liveness,
```
jiexi, line 751: `liveness` should only be set if the networkClientId's chainId matches the globally selected chainId

src/SmartTransactionsController.ts
```diff
@@ -111,9 +118,9 @@ export default class SmartTransactionsController extends BaseController<
 
     this.defaultConfig = {
       interval: DEFAULT_INTERVAL,
-      chainId: CHAIN_IDS.ETHEREUM,
+      chainId: ChainId.mainnet,
```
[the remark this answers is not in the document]
adonesky1: 🤔 yeah don't know why I did this. Will revert.

src/SmartTransactionsController.ts
adonesky1: Its kindof annoying but if I don't add the defaults here and then pass then both as undefined it will override the defaults in `#updateSmartTransaction`.

src/SmartTransactionsController.ts
adonesky1: nope 🤔 also not sure why this change was made...

src/SmartTransactionsController.ts
adonesky1: Yeah perhaps. This was from a while ago but I think my thinking here was optimizing for fewer calls to `this.getNetworkClientById` but it doesn't really matter either way.

src/SmartTransactionsController.ts
adonesky1: good catch! yessir!

src/SmartTransactionsController.ts
adonesky1: yep good catch!

adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/237/commits/caf5ac0a59b73823ef6977cb480bf4c3ec2b5b22

src/SmartTransactionsController.ts
adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/237/commits/caf5ac0a59b73823ef6977cb480bf4c3ec2b5b22

src/SmartTransactionsController.ts
adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/237/commits/caf5ac0a59b73823ef6977cb480bf4c3ec2b5b22

src/SmartTransactionsController.ts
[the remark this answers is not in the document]
adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/237/commits/caf5ac0a59b73823ef6977cb480bf4c3ec2b5b22

src/SmartTransactionsController.test.ts
adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/237/commits/caf5ac0a59b73823ef6977cb480bf4c3ec2b5b22

src/SmartTransactionsController.ts
```diff
@@ -126,13 +133,28 @@ export default class SmartTransactionsController extends BaseController<
...
+          },
+          [CHAIN_IDS.GOERLI]: {
+            approvalTxFees: undefined,
+            tradeTxFees: undefined,
+          },
+        },
       },
     };
-
+    this.setIntervalLength(this.config.interval || DEFAULT_INTERVAL);
```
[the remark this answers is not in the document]
adonesky1: this can/should be removed but I need to move the `this.initialize()` call above. Without moving this we were always using the DEFAULT_INTERVAL. done here: https://github.com/MetaMask/smart-transactions-controller/pull/237/commits/8e2ee0de68987a45c13375a2770300251bf08e7f

src/SmartTransactionsController.ts
```diff
@@ -43,37 +42,43 @@ export const DEFAULT_INTERVAL = SECOND * 5;
...
+export default class SmartTransactionsController extends PollingControllerV1<
   SmartTransactionsControllerConfig,
   SmartTransactionsControllerState
 > {
   public timeoutHandle?: NodeJS.Timeout;
 
   private getNonceLock: any;
 
-  public ethersProvider: any;
+  public ethQuery: EthQuery;
```
jiexi, line 74: probably should be private. nbd deal

src/SmartTransactionsController.test.ts
```diff
@@ -359,12 +425,19 @@ describe('SmartTransactionsController', () => {
...
+        selectedNetworkClientId: 'networkClientId',
+        networkConfigurations: {},
+        networksMetadata: {},
+      } as NetworkState);
       expect(updateSmartTransactionsSpy).not.toHaveBeenCalled();
     });
   });
 
   describe('updateSmartTransactions', () => {
+    // TODO rewrite this test... updateSmartTransactions is getting called via the checkPoll method which is called whenever state is updated.
```
jiexi, line 439: yeah... have run into issues like this as well when testing these polling controllers. Could try stubbing the prototype before instantiating the controller. Up to you if you want to address here as we found the test in this state

src/SmartTransactionsController.test.ts
```diff
@@ -451,6 +524,36 @@ describe('SmartTransactionsController', () => {
...
+      const tradeTx = createUnsignedTransaction(goerliChainIdDec);
+      const approvalTx = createUnsignedTransaction(goerliChainIdDec);
+      const getFeesApiResponse = createGetFeesApiResponse();
+      nock(API_BASE_URL)
+        .post(`/networks/${goerliChainIdDec}/getFees`)
+        .reply(200, getFeesApiResponse);
+
+      expect(
+        smartTransactionsController.state.smartTransactionsState.feesByChainId,
+      ).toStrictEqual(defaultState.smartTransactionsState.feesByChainId);
```
jiexi, line 538: is this needed? I guess the reason why you're doing this is because you're not confident that outside polling hasn't changed this for us?

src/SmartTransactionsController.test.ts
```diff
@@ -744,4 +889,137 @@ describe('SmartTransactionsController', () => {
...
+    let clock: sinon.SinonFakeTimers;
+    beforeEach(() => {
+      clock = sinon.useFakeTimers();
+    });
+
+    afterEach(() => {
+      clock.restore();
+    });
+
+    it('starts and stops calling smart transactions batch status api endpoint with the correct chainId at the polling interval', async () => {
```
jiexi, line 903: would it be to tedious to use matchers and `expect(spy).toHaveBeenNthCalledWith()` instead of `mockClear`?

src/SmartTransactionsController.test.ts
adonesky1: yeah def not strictly needed, I just tend to add these extra assertions to make sure that the later assertions are still testing what they think they are.

src/SmartTransactionsController.test.ts
adonesky1: sure that works!

adonesky1: done here: https://github.com/MetaMask/smart-transactions-controller/pull/237/commits/2b08ccd27a429deaad6d7e318a613ab5b750626b

src/SmartTransactionsController.ts
adonesky1: done here https://github.com/MetaMask/smart-transactions-controller/pull/237/commits/2b08ccd27a429deaad6d7e318a613ab5b750626b

jiexi (approved):
🚀

adonesky1:
Going to wait for a review from @dan437 before merging

PR #271, opened 2024-02-01
Add support for more transactions, emit events

dan437:
## **Description**
This PR does 2 things:
- Adds support for more transactions
- Emits events, so it's easier to understand what's going on with a transaction status

## **Testing**
- Make a build with this PR, include it in the MM extension and make sure Smart Swaps still work

### DIFF
--- a/src/SmartTransactionsController.ts
+++ b/src/SmartTransactionsController.ts
@@ -590,7 +590,7 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
 
   async getFees(
     tradeTx: UnsignedTransaction,
-    approvalTx: UnsignedTransaction,
+    approvalTx?: UnsignedTransaction,
     { networkClientId }: { networkClientId?: NetworkClientId } = {},
   ): Promise<Fees> {
     const chainId = this.#getChainId({ networkClientId });
--- a/src/SmartTransactionsController.ts
+++ b/src/SmartTransactionsController.ts
@@ -79,7 +79,7 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
 
   private trackMetaMetricsEvent: any;
 
-  private eventEmitter: any;
+  private eventEmitter: EventEmitter;
 
   private getNetworkClientById: NetworkController['getNetworkClientById'];
 
--- a/src/SmartTransactionsController.ts
+++ b/src/SmartTransactionsController.ts
@@ -605,13 +605,12 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
         // If there is an approval tx, the trade tx's nonce is increased by 1.
         nonce: incrementNonceInHex(unsignedApprovalTransactionWithNonce.nonce),
       };
-    } else {
+    } else if (tradeTx.nonce) {
       unsignedTradeTransactionWithNonce = tradeTx;
-      if (!unsignedTradeTransactionWithNonce.nonce) {
-        unsignedTradeTransactionWithNonce = await this.addNonceToTransaction(
-          unsignedTradeTransactionWithNonce,
-        );
-      }
+    } else {
+      unsignedTradeTransactionWithNonce = await this.addNonceToTransaction(
+        tradeTx,
+      );
     }
     transactions.push(unsignedTradeTransactionWithNonce);
     const data = await this.fetch(getAPIRequestURL(APIType.GET_FEES, chainId), {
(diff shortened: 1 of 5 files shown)


src/SmartTransactionsController.ts
```diff
@@ -590,9 +606,12 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
         nonce: incrementNonceInHex(unsignedApprovalTransactionWithNonce.nonce),
       };
     } else {
-      unsignedTradeTransactionWithNonce = await this.addNonceToTransaction(
-        tradeTx,
-      );
+      unsignedTradeTransactionWithNonce = tradeTx;
+      if (!unsignedTradeTransactionWithNonce.nonce) {
```
infiniteflower, line 610: This reads a little odd to me. Var name is `withNonce` but it might not exist on the object.

src/SmartTransactionsController.ts
```diff
@@ -590,9 +606,12 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
+        unsignedTradeTransactionWithNonce = await this.addNonceToTransaction(
+          unsignedTradeTransactionWithNonce,
+        );
+      }
```
infiniteflower, line 614: Not a huge deal, but this reads a little odd to me. Var name is `withNonce` but it might not exist on the object. Is it possible to do the following: 

```suggestion
      if (!unsignedTradeTransactionWithNonce.nonce) {
        unsignedTradeTransactionWithNonce = await this.addNonceToTransaction(
          tradeTx,
        );
      }
```

src/SmartTransactionsController.ts
```diff
@@ -670,13 +691,16 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
+      ? await this.getNonceLock(txParams?.from)
+      : undefined;
     try {
-      const nonce = hexlify(nonceLock.nextNonce);
+      const nonce = nonceLock ? hexlify(nonceLock.nextNonce) : txParams.nonce;
       if (txParams && !txParams?.nonce) {
         txParams.nonce = nonce;
       }
-      const { nonceDetails } = nonceLock;
+      const nonceDetails = nonceLock ? nonceLock.nonceDetails : {};
```
wachunei, line 703: This seems like can be refactored in a simpler if case:

```ts
    const requiresNonce = !txParams.nonce;
    let nonce;
    let nonceLock;
    let nonceDetails = {};

    
    if (requiresNonce) {
      nonceLock = await this.getNonceLock(txParams?.from)
      nonce = hexlify(nonceLock.nextNonce)
      nonceDetails = nonceLock.nonceDetails
      if (txParams) txParams.nonce ??= nonce;
    }
  ```
  
  I'm not familiar with the returned types though

**Review by infiniteflower — COMMENTED** [review]
Just had some small comments

File: src/SmartTransactionsController.ts (hunk 3 above)
Comment (dan437) [reply src/SmartTransactionsController.ts:703]: Will simplify it.

File: src/SmartTransactionsController.ts (hunk 2 above)
Comment (dan437) [reply src/SmartTransactionsController.ts:614]: Agreed, but we still need to set the `tradeTx` into `unsignedTradeTransactionWithNonce` if it has a nonce.

I've done this:
```
else if (tradeTx.nonce) {
      unsignedTradeTransactionWithNonce = tradeTx;
    } else {
      unsignedTradeTransactionWithNonce = await this.addNonceToTransaction(
        tradeTx,
      );
    }
```

**Review by infiniteflower — DISMISSED** [review]

File: src/SmartTransactionsController.ts [hunk 4]
```diff
@@ -532,6 +542,12 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
     >;
 
     Object.entries(data).forEach(([uuid, stxStatus]) => {
+      const transactionHash = stxStatus?.minedHash;
+      this.eventEmitter.emit(`${uuid}:status`, stxStatus);
```
mcmire, line 546: Where will this event be used if `eventEmitter` is private?

src/SmartTransactionsController.ts
```diff
@@ -77,6 +79,8 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
 
   private trackMetaMetricsEvent: any;
 
+  private eventEmitter: any;
```
mcmire, line 82: Can we give this property a type? Perhaps:

``` suggestion
  private eventEmitter: EventEmitter;

mcmire (commented):
Hello, I suspect there are plans here for events, but I've seen similar events in TransactionController as those you're adding here, so I'm curious how we intend to use them? Thanks.

src/SmartTransactionsController.ts
dan437: Extension was able to use it. More here about the `events` node module: https://www.npmjs.com/package/events

infiniteflower (approved):

PR #316, opened 2024-04-15
Add a new function "wipeSmartTransactions", replace Goerli with Sepolia

dan437:
## Description
This PR does 2 things:
- Adds a new function "wipeSmartTransactions"
- Replaces Goerli with Sepolia

The "wipeSmartTransactions" function is needed when a user wants to clear their Activity in MM. It's done similarly to what's implemented in the TransactionController.

### DIFF
--- a/src/SmartTransactionsController.ts
+++ b/src/SmartTransactionsController.ts
@@ -57,7 +57,7 @@ export type SmartTransactionsControllerConfig = BaseConfig & {
   interval: number;
   clientId: string;
   chainId: Hex;
-  supportedChainIds: string[];
+  supportedChainIds: Hex[];
 };
 
 type FeeEstimates = {
@@ -888,5 +888,57 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
         txHash.toLowerCase()
       );
     });
+  }
+
+  #wipeSmartTransactionsPerChainId(
+    chainId: Hex,
+    selectedAddressLowerCase: string,
+  ): void {
+    const { smartTransactions } = this.state.smartTransactionsState;
+    const smartTransactionsForSelectedChain: SmartTransaction[] =
+      smartTransactions?.[chainId];
+    if (
+      !smartTransactionsForSelectedChain ||
+      smartTransactionsForSelectedChain.length === 0
+    ) {
+      return;
+    }
+    const newSmartTransactionsForSelectedChain =
+      smartTransactionsForSelectedChain.filter(
+        (smartTransaction: SmartTransaction) =>
+          smartTransaction.txParams?.from !== selectedAddressLowerCase,
+      );
+    this.update({
+      smartTransactionsState: {
+        ...this.state.smartTransactionsState,
+        smartTransactions: {
+          ...smartTransactions,
+          [chainId]: newSmartTransactionsForSelectedChain,
+        },
+      },
+    });
+  }
+
+  wipeSmartTransactions(
+    selectedAddress?: string,
+    ignoreNetwork?: boolean,
+  ): void {
+    if (!selectedAddress) {
+      return;
+    }
+    const selectedAddressLowerCase = selectedAddress.toLowerCase();
+    if (ignoreNetwork) {
+      this.#wipeSmartTransactionsPerChainId(
+        this.config.chainId,
+        selectedAddressLowerCase,
+      );
+    } else {
+      this.config.supportedChainIds.forEach((chainId) => {
+        this.#wipeSmartTransactionsPerChainId(
+          chainId,
+          selectedAddressLowerCase,
+        );
+      });
+    }
   }
 }
--- a/src/SmartTransactionsController.ts
+++ b/src/SmartTransactionsController.ts
@@ -1,7 +1,7 @@
 // eslint-disable-next-line import/no-nodejs-modules
 import { hexlify } from '@ethersproject/bytes';
 import type { BaseConfig, BaseState } from '@metamask/base-controller';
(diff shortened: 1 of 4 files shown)


src/SmartTransactionsController.ts
```diff
@@ -889,4 +889,56 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
+        },
+      },
+    });
+  }
+
+  wipeSmartTransactions(
+    selectedAddress?: string,
+    ignoreNetwork?: boolean,
+  ): void {
+    if (!selectedAddress) {
```
matthewwalsh0, line 926: Is this intentional? In the `TransactionController`, the `selectedAddress` is optional since it's an optional filter meaning if it's not provided, transactions for all accounts are removed.

src/SmartTransactionsController.ts
```diff
@@ -889,4 +889,56 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
+        smartTransactions: {
+          ...smartTransactions,
+          [chainId]: newSmartTransactionsForSelectedChain,
+        },
+      },
+    });
+  }
+
+  wipeSmartTransactions(
+    selectedAddress?: string,
```
matthewwalsh0, line 923: Do we need to call this `selectedAddress` or would `address` be simpler since whether it is selected could be domain knowledge this method doesn't need to know?

matthewwalsh0, line 923: For arguments like this with many optional arguments, and booleans especially, I prefer object arguments so it's more readable when used, and easy to add additional filters in future.

src/SmartTransactionsController.ts
```diff
@@ -889,4 +889,56 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
+      return;
+    }
+    const selectedAddressLowerCase = selectedAddress.toLowerCase();
+    if (ignoreNetwork) {
+      this.#wipeSmartTransactionsPerChainId(
+        this.config.chainId,
+        selectedAddressLowerCase,
+      );
+    } else {
+      this.config.supportedChainIds.forEach((chainId) => {
```
matthewwalsh0, line 936: Rather than looping through every supported chain ID, should we instead iterate over the state itself and the chain IDs in it?

Could this way introduce gaps if the config is changed to support a smaller set of chain IDs and old transactions exist with chain IDs no longer supported?

src/SmartTransactionsController.ts
```diff
@@ -889,4 +889,56 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
+
+  wipeSmartTransactions(
+    selectedAddress?: string,
+    ignoreNetwork?: boolean,
+  ): void {
+    if (!selectedAddress) {
+      return;
+    }
+    const selectedAddressLowerCase = selectedAddress.toLowerCase();
+    if (ignoreNetwork) {
```
matthewwalsh0, line 930: Does using this condition and helper method here complicate the method a little and make it more rigid if additional filters are added in future?

Would it be simpler and more extensible to have a single loop going through a flattened list of all smart transactions and filtering as needed, then the map can be reassembled at the end with a `#groupTransactionsByChainId` for example?

You could even use a helper to get the flattened list such as `#getAllSmartTransactions()`.

src/SmartTransactionsController.ts
```diff
@@ -889,4 +889,56 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
       );
     });
   }
+
+  #wipeSmartTransactionsPerChainId(
```
matthewwalsh0, line 893: Minor and may not be the case in this file, but typically we try and group public methods together at the top with private methods below them, for easier visibility.

src/constants.ts
```diff
@@ -15,3 +15,8 @@ export enum MetaMetricsEventName {
 export enum MetaMetricsEventCategory {
   Transactions = 'Transactions',
 }
+
+export enum NetworkClientId {
```
matthewwalsh0, line 19: Does this already exist in the `NetworkController` since it generates the network client IDs?

src/constants.ts
```diff
@@ -1,7 +1,7 @@
 export const API_BASE_URL = 'https://transaction.api.cx.metamask.io';
 export const CHAIN_IDS = {
```
matthewwalsh0, line 2: It's an awkward one to maintain but the `NetworkController` has a list of chain IDs, plus the `TransactionController`, plus the extension itself.

I wonder if it's worth using a single source of truth in future such as the `NetworkController`.

src/SmartTransactionsController.ts
dan437: Yes, I have this check and early return as part of defensive programming, just in case the value would be falsy (e.g. an empty string), since this function will be called from a .js file, which is dynamically typed. However, I will make a tiny change to make this param mandatory.

src/SmartTransactionsController.ts
dan437: Yes, I usually use object arguments. Will do it here as well.

src/SmartTransactionsController.ts
dan437: Which additional filters are you thinking about? I don't think this function will change anytime soon and wiping smart transactions per chainId seems simple enough.

src/constants.ts
dan437: That would be nice if all controllers and extension / mobile use one source of truth for chainIds. Will get it from the NetworkController.

matthewwalsh0 (approved):

PR #353, opened 2024-06-04
fix: Improve state management

dan437:
## Description
We [fixed this issue](https://github.com/MetaMask/smart-transactions-controller/pull/347), eliminating an edge case related to network switching where we were unnecessarily polling for transaction status.

This PR improves state management to ensure unique smart transactions, and includes code to fix this issue for a small subset of users who have encountered this issue. The code to fix the issue for this subset of users will be removed in a future version once we have confirmed this is resolved.

### DIFF
--- a/src/SmartTransactionsController.test.ts
+++ b/src/SmartTransactionsController.test.ts
@@ -462,10 +462,6 @@ describe('SmartTransactionsController', () => {
     });
 
     it('calls "ensureUniqueSmartTransactions" on network change if it is a new chainId', () => {
-      const ensureUniqueSmartTransactionsSpy = jest.spyOn(
-        smartTransactionsController,
-        'ensureUniqueSmartTransactions',
-      );
       const { smartTransactionsState } = smartTransactionsController.state;
       const smartTransactionsForChainId = createStateAfterSuccess();
       smartTransactionsForChainId.push({
@@ -496,7 +492,6 @@ describe('SmartTransactionsController', () => {
       const uniqueSmartTransactionsForChainId =
         smartTransactionsController.state.smartTransactionsState
           .smartTransactions[ChainId.mainnet];
-      expect(ensureUniqueSmartTransactionsSpy).toHaveBeenCalled();
       expect(uniqueSmartTransactionsForChainId).toHaveLength(1);
       expect(uniqueSmartTransactionsForChainId).toStrictEqual([
         smartTransactionsForChainId[0],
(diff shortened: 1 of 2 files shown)


src/SmartTransactionsController.test.ts
```diff
@@ -460,6 +460,66 @@ describe('SmartTransactionsController', () => {
       } as NetworkState);
       expect(checkPollSpy).toHaveBeenCalled();
     });
+
+    it('calls "ensureUniqueSmartTransactions" on network change if it is a new chainId', () => {
```
mcmire, line 464: I see that we check that this method is called, but we don't test what this method does. Do we want to add tests for that method in particular? Or, if the method isn't intended to be called on its own, should it be made private and then we test the behavior of the method in this test?

dan437: So, there are tests to check that the function removed duplicities properly: https://github.com/MetaMask/smart-transactions-controller/pull/353/files#diff-750ce12ce9485eea01d3ee2172410c52fdd60e40cb3eef6c7b0909b8e8999284R496-R503

src/SmartTransactionsController.ts
```diff
@@ -238,6 +243,35 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
+    }
+    const uniqueUUIDs = new Set();
+    const uniqueSmartTransactionsForChainId = [];
+    for (const transaction of smartTransactionsForChainId) {
+      if (!uniqueUUIDs.has(transaction.uuid)) {
+        uniqueUUIDs.add(transaction.uuid);
+        uniqueSmartTransactionsForChainId.push(transaction);
+      }
+    }
+    this.update({
```
dbrans, line 264:
```suggestion
  const uniqueSmartTransactionsForChainId = Array.from(
    new Map(smartTransactionsForChainId.map(
      (transaction) => [transaction.uuid, transaction])
      ).values()
  );
  this.update({
```

src/SmartTransactionsController.test.ts
mcmire: Does the method need to be public then, at least? Do we anticipate consumers needing to call it?

dbrans (approved):

src/SmartTransactionsController.test.ts
dan437: It's only a private function and will be removed in one of the upcoming releases.

dan437: I've marked the function as private in the code to make it clearer.

mcmire: Did you forget to push? :)

mcmire (approved):
Great, LGTM!

PR #374, opened 2024-07-09
Update metrics, refactoring

dan437:
## Description
This PR updates metrics for 2 events and includes a bit of refactoring, which was needed to make the events work for non-swaps transactions.

### DIFF
--- a/src/SmartTransactionsController.ts
+++ b/src/SmartTransactionsController.ts
@@ -34,7 +34,6 @@ import {
   calculateStatus,
   generateHistoryEntry,
   getAPIRequestURL,
-  getStxProcessingTime,
   handleFetch,
   incrementNonceInHex,
   isSmartTransactionCancellable,
@@ -42,6 +41,7 @@ import {
   replayHistory,
   snapshotFromTxMeta,
   getTxHash,
+  getSmartTransactionMetricsProperties,
 } from './utils';
 
 const SECOND = 1000;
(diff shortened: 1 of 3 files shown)


src/SmartTransactionsController.ts
```diff
@@ -363,13 +345,47 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
+    }
+    this.update({
+      smartTransactionsState: {
+        ...smartTransactionsState,
+        smartTransactions: {
+          ...smartTransactionsState.smartTransactions,
+          [chainId]: smartTransactionsState.smartTransactions[chainId].map(
+            (item, index) => {
+              return index === currentIndex
+                ? { ...item, ...smartTransaction }
```
dbrans, line 379: Just a heads-up: this method updates only the properties present in the provided `smartTransaction`. This merge-like behavior might be surprising for some users. Case in point, looking at line 593 in this same PR, if "merge" were the expected bahaviour, you could just write `{confirmed: true}`.

src/SmartTransactionsController.ts
```diff
@@ -363,13 +345,47 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
+    if (currentIndex === -1) {
+      return; // Smart transaction not found, don't update anything.
+    }
+    this.update({
+      smartTransactionsState: {
+        ...smartTransactionsState,
+        smartTransactions: {
+          ...smartTransactionsState.smartTransactions,
+          [chainId]: smartTransactionsState.smartTransactions[chainId].map(
+            (item, index) => {
```
dbrans, line 377: Hey, consider naming the item parameter something more specific like existingSmartTransaction to make it clearer.

dbrans (dismissed):

src/SmartTransactionsController.ts
dan437: We could only pass `{ confirmed: true }` on line 593, but the `#updateSmartTransaction` function expects the `SmartTransaction` type for the `smartTransaction` input param, which also requires `uuid`. I believe that for simplicity it's easier to just pass the whole `smartTransaction` to be updated in state.

dbrans (approved):

PR #384, opened 2024-07-12
adapt to eip-1193 provider changes

### LINKED ISSUE #306: Adapt to EIP-1193 provider changes
After [`SafeEventEmitterProvider` is updated to support EIP-1193](https://github.com/MetaMask/core/issues/4095) and a new version of `@metamask/eth-json-rpc-provider` is released, we should adapt to the changes:

- We should add `@metamask/eth-json-rpc-provider` as a dependency.
- This controller uses `Provider` from `@metamask/eth-query`. That's incorrect; we should use `SafeEventEmitterProvider` instead.
- Calling `sendAsync` will be deprecated; we should use `request` instead.
  - The only references to `sendAsync` are in tests, where we pass a mock provider. We should pass a provider with a `request` method instead.

kanthesha:
Calling sendAsync will be deprecated; we should use request instead.

Fixes #306

### DIFF
--- a/src/SmartTransactionsController.test.ts
+++ b/src/SmartTransactionsController.test.ts
@@ -342,13 +342,9 @@ const defaultState = {
   },
 };
 
-const mockProvider = {
-  request: jest.fn(),
-};
-
 const mockProviderConfig = {
   chainId: ChainId.mainnet,
-  provider: mockProvider,
+  provider: getFakeProvider(),
   type: NetworkType.mainnet,
   ticker: 'ticker',
 };


cryptodev-2s (dismissed):
LGTM!

src/SmartTransactionsController.test.ts
```diff
@@ -343,7 +343,7 @@ const defaultState = {
 };
 
 const mockProvider = {
-  sendAsync: jest.fn(),
+  request: jest.fn(),
 };
```
mikesposito, line 347: Should `mockProvider` be substituted with `getFakeProvider()` as well, or as an alternative use `mockProvider` instead of `getFakeProvider()` in the other places?

mikesposito: If `mockProvider` is enough for what we need, it would probably keep tests simpler to understand

mikesposito (approved):
Looks good!

PR #386, opened 2024-07-15
Save new event props to a newly created smart transaction, use both `properties` and `sensitiveProperties` for events

dan437:
## Description
Save new event props to a newly created smart transaction, so they can be used in anonymised metrics (only if a user opted into them).


dbrans (approved):

dbrans:
Nice work. LGTM!

PR #395, opened 2024-07-29
upgrade NetworkController to V20

### LINKED ISSUE #391: Upgrade `@metamask/network-controller` from v19 to v20
`@metamask/network-controller` 20.0.0 removed `providerConfig` from state, which means that we now need to use `selectedNetworkClientId` to access the chain ID for the currently selected network.

This is a prerequisite for upgrading to BaseController v2.

kanthesha:
Upgrade the @metamask/network-controller to v20. [@metamask/network-controller 20.0.0](https://github.com/MetaMask/core/blob/main/packages/network-controller/CHANGELOG.md#2000) removed providerConfig from state, which means that we now need to use selectedNetworkClientId to access the chain ID for the currently selected network. And we no longer need to take provider, instead get this from the network client object.

* Fixes #391

### DIFF
--- a/src/SmartTransactionsController.ts
+++ b/src/SmartTransactionsController.ts
@@ -189,10 +189,11 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
     this.initializeSmartTransactionsForChainId();
     this.#ensureUniqueSmartTransactions();
 
-    onNetworkStateChange((networkState) => {
-      const { selectedNetworkClientId } = networkState;
-      const networkClient = this.getNetworkClientById(selectedNetworkClientId);
-      const { chainId } = networkClient.configuration;
+    onNetworkStateChange(({ selectedNetworkClientId }) => {
+      const {
+        configuration: { chainId },
+        provider,
+      } = this.getNetworkClientById(selectedNetworkClientId);
       const isNewChainId = chainId !== this.config.chainId;
       this.configure({ chainId });
       this.initializeSmartTransactionsForChainId();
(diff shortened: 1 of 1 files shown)


src/SmartTransactionsController.ts
```diff
@@ -192,16 +189,18 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
       const isNewChainId = chainId !== this.config.chainId;
       this.configure({ chainId });
       this.initializeSmartTransactionsForChainId();
       if (isNewChainId) {
         this.#ensureUniqueSmartTransactions();
       }
       this.checkPoll(this.state);
-      this.ethQuery = new EthQuery(provider);
+      this.ethQuery = new EthQuery(networkClient.provider);
     });
```
cryptodev-2s, line 204: Nit: Can we make it this way instead. the only difference is less rows ?
```suggestion
 onNetworkStateChange(({ selectedNetworkClientId }) => {
    const {
      configuration: { chainId },
      provider,
    } = this.getNetworkClientById(selectedNetworkClientId);
    const isNewChainId = chainId !== this.config.chainId;
    this.configure({ chainId });
    this.initializeSmartTransactionsForChainId();
    if (isNewChainId) {
      this.#ensureUniqueSmartTransactions();
    }
    this.checkPoll(this.state);
    this.ethQuery = new EthQuery(provider);
  });
```

cryptodev-2s (commented):
Looks good to me, there's just one minor change proposed

cryptodev-2s (approved):
LGTM!

Issue #385, opened 2024-07-12
Update to BaseControllerV2


Gudahtt:
Update this controller to use the new BaseController class (here: https://github.com/MetaMask/core/blob/448f8f02378be606beecdc436c5f5cc59dd906c9/packages/base-controller/src/BaseControllerV2.ts).

The BaseControllerV1 class currently in-use is deprecated and no longer maintained. It will soon be removed. This is one of the last two instances, and is currently blocking some improvements to extension and mobile state management that require the use of BaseControllerV1.

## Acceptance Criteria

- A constant called `controllerName` exists which holds the name of the controller.
- A constant called `controllerMetadata` exists which holds the metadata for the state.
- The default state object is extracted out of the constructor to a function called `getDefaultSmartTransactionsControllerState`.
- `SmartTransactionsController` inherits from `StaticIntervalPollingController` instead of `StaticIntervalPollingControllerV1`.
  - The `SmartTransactionsController` constructor signature is updated to take a single options bag instead of three arguments, with options in `config` merged into options not in `state` or `config`.
  - The constructor no longer takes `onNetworkStateChange`, but subscribes to `NetworkController:stateChange`.
    - Note that this requires upgrading `@metamask/network-controller` to v20 first.
  - The constructor no longer takes `getNetworkClientById`, but calls this through the messenger instead.
  - The controller no longer subscribes to itself by calling `subscribe` on itself, but on the messenger instead.
- An `EventEmitter` is no longer to emit events, but the messenger is used instead.
- Supporting types exist and/or are updated.
- - The `SmartTransactionsControllerConfig` type no longer exists, and is renamed to `SmartTransactionsControllerOptions`.
  - The `SmartTransactionsControllerGetStateAction` and `SmartTransactionsControllerStateChangeEvent` types exist.
  - The `SmartTransactionsControllerSmartTransactionEvent` type exists.
  - The `SmartTransactionsControllerActions` type exists (it is simply `never`).
  - The `SmartTransactionsControllerEvents` type exists, and includes `SmartTransactionsControllerSmartTransactionEvent`.
  - The `AllowedActions` type exists, and includes `NetworkControllerGetNetworkClientByIdAction`.
  - and `AllowedEvents` type exists, and includes `NetworkControllerStateChangeEvent`.
  - The `SmartTransactionsControllerMessenger` type exists and expects `NetworkController:getNetworkClientById` and `NetworkController:stateChange` to be allowed.
[...]

PR #397, opened 2024-08-01
feat: update controller to inherit from BaseControllerV2

cryptodev-2s:
SmartTransactionController has been updated to inherit from `StaticIntervalPollingController` instead of `StaticIntervalPollingControllerV2`

Fixes [#385](https://github.com/MetaMask/smart-transactions-controller/issues/385)


src/SmartTransactionsController.ts
```diff
@@ -47,165 +48,237 @@ import {
...
+ * The name of the {@link SmartTransactionsController}
+ */
+const controllerName = 'SmartTransactionsController';
+
+const controllerMetadata = {
+  smartTransactionsState: {
+    persist: false,
+    anonymous: true,
+  },
 };
```
mikesposito, line 65: Hmm, putting the whole controller state under the `smartTransactionsState` forces us to apply the same metadata to all of it. I know that this was already here before this PR, but thoughts on removing this key and putting the entire state as root object?

src/SmartTransactionsController.ts
```diff
@@ -47,165 +48,237 @@ import {
...
+  readonly #getNonceLock: any;
 
-  private ethQuery: EthQuery | undefined;
+  #ethQuery: EthQuery | undefined;
 
   public confirmExternalTransaction: any;
 
   public getRegularTransactions: (
     options?: GetTransactionsOptions,
   ) => TransactionMeta[];
```
mikesposito, line 201: nit: do you think we can lower the visibility of these public variables?

src/SmartTransactionsController.ts
```diff
@@ -364,41 +429,36 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
-        },
-      },
+
+    this.update((state) => {
+      const currentTransaction = cloneDeep(
+        state.smartTransactionsState.smartTransactions[chainId][currentIndex],
+      );
+      state.smartTransactionsState.smartTransactions[chainId][currentIndex] =
+        Object.assign(currentTransaction, smartTransaction);
     });
```
mikesposito, line 455: nit: do you think we can make this more idimatic?
```suggestion
      state.smartTransactionsState.smartTransactions[chainId][currentIndex] = {
        ...state.smartTransactionsState.smartTransactions[chainId][
          currentIndex
        ],
        ...smartTransaction,
      };
    });
```

src/SmartTransactionsController.ts
```diff
@@ -47,165 +48,237 @@ import {
...
+export type AllowedActions = NetworkControllerGetNetworkClientByIdAction;
+
+export type SmartTransactionsControllerStateChangeEvent =
+  ControllerStateChangeEvent<
+    typeof controllerName,
+    SmartTransactionsControllerState
+  >;
+
+export type SmartTransactionsControllerSmartTransactionEvent = {
+  type: 'SmartTransactionsController:smartTransaction';
```
mikesposito, line 139: nit: Do you think we can make this event name more meaningful by describing what happened with the smart transaction? Perhaps something like `SmartTransactionsController:smartTransactionCreated`?

src/SmartTransactionsController.ts
```diff
@@ -718,38 +775,35 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
+    const data = await this.#fetch(
+      getAPIRequestURL(APIType.GET_FEES, chainId),
+      {
+        method: 'POST',
+        body: JSON.stringify({
+          txs: transactions,
+        }),
+      },
+    );
+    let approvalTxFees: IndividualTxFees | null = null;
```
mikesposito, line 787: nit:
```suggestion
    let approvalTxFees: IndividualTxFees | null;
```

src/SmartTransactionsController.ts
```diff
@@ -840,9 +894,14 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
   }: { networkClientId?: NetworkClientId } = {}): Hex {
-    return networkClientId
-      ? this.getNetworkClientById(networkClientId).configuration.chainId
-      : this.config.chainId;
+    if (networkClientId) {
+      return this.messagingSystem.call(
+        'NetworkController:getNetworkClientById',
+        networkClientId,
+      ).configuration.chainId;
+    }
```
mikesposito, line 902: Do you think this is still necessary since we change `this.chainId` everytime `NetworkController` emits a state update?

src/SmartTransactionsController.ts
cryptodev-2s: This will assume there will be a need for a migration script we can maybe create an issue for this and address this change in a separate ticket

src/SmartTransactionsController.ts
cryptodev-2s: This is what I have done in the beginning check the first commit 
Then the ci security check flagged this as "Prototype-polluting assignment"

src/SmartTransactionsController.ts
cryptodev-2s: They should actually be null and not undefined as undefined is not serialisable on state

src/SmartTransactionsController.ts
MajorLift: To fix the "prototype-polluting assignment" error, we have the `isSafeDynamicKey` utility in controller-utils (see https://github.com/MetaMask/core/pull/4041).

```suggestion
      const currentTransaction = cloneDeep(
        state.smartTransactionsState.smartTransactions[chainId][currentIndex],
      );
      if (
        [chainId, currentIndex].every((key) => isSafeDynamicKey(String(key)))
      ) {
        state.smartTransactionsState.smartTransactions[chainId][currentIndex] =
          Object.assign(currentTransaction, smartTransaction);
      }
    });
```

src/SmartTransactionsController.ts
```diff
@@ -718,38 +775,35 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
+      getAPIRequestURL(APIType.GET_FEES, chainId),
+      {
+        method: 'POST',
+        body: JSON.stringify({
+          txs: transactions,
+        }),
+      },
+    );
+    let approvalTxFees: IndividualTxFees | null = null;
+    let tradeTxFees: IndividualTxFees | null;
```
MajorLift: Should both variables be initialized to `null`?

```suggestion
    let approvalTxFees: IndividualTxFees | null = null;
    let tradeTxFees: IndividualTxFees | null = null;
```

src/SmartTransactionsController.ts
MajorLift: I believe we'll still need that logic since `#getChainId` is intended to return the corresponding `chainId` for any `networkClientId`, not just the `selectedNetworkClientId`.

src/SmartTransactionsController.ts
MajorLift, line 1: Looks like we could remove the `any`s in the `submitSignedTransactions` method.

```diff
diff --git a/src/SmartTransactionsController.ts b/src/SmartTransactionsController.ts
index 221d019..07d3b60 100644
--- a/src/SmartTransactionsController.ts
+++ b/src/SmartTransactionsController.ts
@@ -12,7 +12,10 @@ import type {
   NetworkControllerStateChangeEvent,
 } from '@metamask/network-controller';
 import { StaticIntervalPollingController } from '@metamask/polling-controller';
-import type { TransactionMeta } from '@metamask/transaction-controller';
+import type {
+  TransactionMeta,
+  TransactionParams,
+} from '@metamask/transaction-controller';
 import { TransactionStatus } from '@metamask/transaction-controller';
 import { BigNumber } from 'bignumber.js';
 import cloneDeep from 'lodash/cloneDeep';
@@ -823,8 +826,8 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
   }: {
     signedTransactions: SignedTransaction[];
     signedCanceledTransactions: SignedCanceledTransaction[];
-    transactionMeta?: any;
-    txParams?: any;
+    transactionMeta?: TransactionMeta;
+    txParams?: TransactionParams;
     networkClientId?: NetworkClientId;
   }) {
     const chainId = this.#getChainId({ networkClientId });
@@ -850,7 +853,7 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
       console.error('provider error', error);
     }
 
-    const requiresNonce = !txParams.nonce;
+    const requiresNonce = !txParams?.nonce;
     let nonce;
     let nonceLock;
     let nonceDetails = {};
@@ -880,7 +883,7 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
           uuid: submitTransactionResponse.uuid,
           txHash: submitTransactionResponse.txHash,
           cancellable: true,
-          type: transactionMeta?.type || 'swap',
+          type: transactionMeta?.type ?? 'swap',
         },
         { chainId, ethQuery },
       );

```

MajorLift, line 1: We also might be able to remove the `any`s in the constructor options.

```diff
diff --git a/package.json b/package.json
index c3f3e28..982d0ae 100644
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "@metamask/eth-json-rpc-provider": "^4.1.0",
     "@metamask/eth-query": "^4.0.0",
     "@metamask/network-controller": "^20.0.0",
+    "@metamask/nonce-tracker": "^6.0.0",
     "@metamask/polling-controller": "^8.0.0",
     "@metamask/transaction-controller": "^34.0.0",
     "bignumber.js": "^9.0.1",
diff --git a/src/SmartTransactionsController.ts b/src/SmartTransactionsController.ts
index 221d019..a45bd6f 100644
--- a/src/SmartTransactionsController.ts
+++ b/src/SmartTransactionsController.ts
@@ -11,8 +11,12 @@ import type {
   NetworkControllerGetNetworkClientByIdAction,
   NetworkControllerStateChangeEvent,
 } from '@metamask/network-controller';
+import type { NonceLock } from '@metamask/nonce-tracker';
 import { StaticIntervalPollingController } from '@metamask/polling-controller';
-import type { TransactionMeta } from '@metamask/transaction-controller';
+import type {
+  TransactionMeta,
+  TransactionReceipt,
+} from '@metamask/transaction-controller';
 import { TransactionStatus } from '@metamask/transaction-controller';
 import { BigNumber } from 'bignumber.js';
 import cloneDeep from 'lodash/cloneDeep';
@@ -166,9 +170,25 @@ type SmartTransactionsControllerOptions = {
   clientId?: string;
   chainId?: Hex;
   supportedChainIds?: Hex[];
-  getNonceLock: any;
-  confirmExternalTransaction: any;
-  trackMetaMetricsEvent: any;
+  getNonceLock: (address?: string) => Promise<NonceLock>;
+  confirmExternalTransaction: (
+    txMeta: TransactionMeta,
+    transactionReceipt: TransactionReceipt,
+    feePerGas?: string,
+  ) => void;
+  trackMetaMetricsEvent: ({
+    event,
+    category,
+    properties,
+    sensitiveProperties,
+  }: {
+    event: MetaMetricsEventName;
+    category: MetaMetricsEventCategory;
+    properties?: ReturnType<typeof getSmartTransactionMetricsProperties>;
+    sensitiveProperties?: ReturnType<
+      typeof getSmartTransactionMetricsSensitiveProperties
+    >;
+  }) => void;
   state?: Partial<SmartTransactionsControllerState>;
   messenger: SmartTransactionsControllerMessenger;
   getTransactions: (options?: GetTransactionsOptions) => TransactionMeta[];
@@ -190,17 +210,17 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
 
   public timeoutHandle?: NodeJS.Timeout;
 
-  readonly #getNonceLock: any;
+  readonly #getNonceLock: SmartTransactionsControllerOptions['getNonceLock'];
 
   #ethQuery: EthQuery | undefined;
 
-  public confirmExternalTransaction: any;
+  public confirmExternalTransaction: SmartTransactionsControllerOptions['confirmExternalTransaction'];
 
   public getRegularTransactions: (
     options?: GetTransactionsOptions,
   ) => TransactionMeta[];
 
-  readonly #trackMetaMetricsEvent: any;
+  readonly #trackMetaMetricsEvent: SmartTransactionsControllerOptions['trackMetaMetricsEvent'];
 
   readonly #getMetaMetricsProps: () => Promise<MetaMetricsProps>;
```

src/SmartTransactionsController.ts
cryptodev-2s: just using cloneDeep and Objet.assign did fix it I believe ?

src/SmartTransactionsController.ts
cryptodev-2s: actually this is an old code I moved the assignment to the statement below

src/SmartTransactionsController.ts
MajorLift: I'm still seeing the warning in this PR.

src/SmartTransactionsController.ts
```diff
@@ -364,41 +431,36 @@
...
-            },
-          ),
-        },
-      },
+
+    this.update((state) => {
+      const currentTransaction = cloneDeep(
+        state.smartTransactionsState.smartTransactions[chainId][currentIndex],
+      );
+      state.smartTransactionsState.smartTransactions[chainId][currentIndex] =
```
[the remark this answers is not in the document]
MajorLift: @cryptodev-2s

src/SmartTransactionsController.ts
```diff
@@ -796,13 +864,13 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
 
-    const requiresNonce = !txParams.nonce;
+    const requiresNonce = !txParams?.nonce;
     let nonce;
     let nonceLock;
     let nonceDetails = {};
 
     if (requiresNonce) {
-      nonceLock = await this.getNonceLock(txParams?.from);
+      nonceLock = await this.#getNonceLock(txParams?.from as string);
```
MajorLift, line 873: `this.#getNonceLock` appears to throw an error if an `address` parameter is not supplied, meaning `txParams` being undefined results in `nonce{,Lock,Details}` all being undefined as well. Therefore, the following change should be equivalent to the original code while resolving the type issues.

```diff
@@ -864,18 +864,16 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
       console.error('provider error', error);
     }
 
-    const requiresNonce = !txParams?.nonce;
+    const requiresNonce = txParams && !txParams.nonce;
     let nonce;
     let nonceLock;
     let nonceDetails = {};
 
     if (requiresNonce) {
-      nonceLock = await this.#getNonceLock(txParams?.from as string);
+      nonceLock = await this.#getNonceLock(txParams.from);
       nonce = hexlify(nonceLock.nextNonce);
       nonceDetails = nonceLock.nonceDetails;
-      if (txParams) {
-        txParams.nonce ??= nonce;
-      }
+      txParams.nonce ??= nonce;
     }
     const submitTransactionResponse = {
       ...data,
```

src/SmartTransactionsController.ts
```diff
@@ -602,13 +674,13 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
             : originalTxMeta;
 
         if (this.#doesTransactionNeedConfirmation(txHash)) {
-          this.confirmExternalTransaction(
-            txMeta,
+          this.#confirmExternalTransaction(
+            txMeta as TransactionMeta,
             transactionReceipt,
-            baseFeePerGas,
+            baseFeePerGas as Hex,
```
MajorLift, line 680: This is less than ideal but does seem to align with the current runtime behavior of this code. Maybe we could add TODO comments so that these as assertions are fixed sooner rather than later.

```diff
@@ -675,8 +675,10 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
 
         if (this.#doesTransactionNeedConfirmation(txHash)) {
           this.#confirmExternalTransaction(
+            // TODO: Replace 'as' assertion with correct typing for `txMeta`
             txMeta as TransactionMeta,
             transactionReceipt,
+            // TODO: Replace 'as' assertion with correct typing for `baseFeePerGas`
             baseFeePerGas as Hex,
           );
         }
```

src/SmartTransactionsController.ts
```diff
@@ -965,7 +1039,9 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
     }
     const addressLowerCase = address.toLowerCase();
     if (ignoreNetwork) {
-      const { smartTransactions } = this.state.smartTransactionsState;
+      const {
+        smartTransactionsState: { smartTransactions },
+      } = this.state;
       Object.keys(smartTransactions).forEach((chainId) => {
         const chainIdHex: Hex = chainId as Hex;
         this.#wipeSmartTransactionsPerChainId({
```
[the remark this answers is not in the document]
MajorLift: nit: The same effect of narrowing a property key type down from `string` can also be achieved with `getKnownPropertyNames` from `@metamask/utils`, but introducing a new dependency just for this single application probably isn't worth it.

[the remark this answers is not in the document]
cryptodev-2s: maybe we can do this later

src/SmartTransactionsController.ts
```diff
@@ -364,41 +434,40 @@
...
-      },
+
+    this.update((state) => {
+      const currentTransaction = cloneDeep(
+        state.smartTransactionsState.smartTransactions[chainId][currentIndex],
+      );
+      if (
+        [chainId, currentIndex].every((key) => isSafeDynamicKey(String(key)))
+      ) {
+        state.smartTransactionsState.smartTransactions[chainId][currentIndex] =
```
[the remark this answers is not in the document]
MajorLift: Hmm looks like this fix didn't get rid of the alert. Could we try something more explicit e.g.

```
if (isSafeDynamicKey(chainId) && isSafeDynamicKey(String(currentIndex))) {
```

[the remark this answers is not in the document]
MajorLift: Not sure if the validation for `currentIndex` will work with the string conversion.

I created a ticket for updating `isSafeDynamicKey` so that it handles all `PropertyKey` subtypes: https://github.com/MetaMask/utils/issues/200

src/SmartTransactionsController.ts
```diff
@@ -364,41 +431,37 @@
...
-              return index === currentIndex
-                ? { ...existingSmartTransaction, ...smartTransaction }
-                : existingSmartTransaction;
-            },
-          ),
-        },
-      },
+
+    this.update((state) => {
+      state.smartTransactionsState.smartTransactions[chainId][currentIndex] = {
```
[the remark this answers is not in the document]
cryptodev-2s: @legobeat any thoughts on how to fix this issue ?

MajorLift (dismissed):
LGTM!

MajorLift (approved):

PR #413, opened 2024-09-03
Update release 12.0.1 CHANGELOG

cryptodev-2s:
This PR updates the release `12.0.1` CHANGELOG as per discussions on this thread https://github.com/MetaMask/smart-transactions-controller/pull/407#discussion_r1727629351

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ## [12.0.1]
-### Removed
-- Remove logic for ensuring uniqueness of smart transactions ([#404](https://github.com/MetaMask/smart-transactions-controller/pull/404))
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -7,9 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ## [12.0.1]
+### Removed
```
Gudahtt, line 10: The "Removed" category is for "removed features", see https://keepachangelog.com/en/1.0.0/
A removed feature is typically a breaking change, as someone might be relying on that feature. But this was changed in a patch release and isn't marked as breaking.

Maybe it's better categorized under "changed"? If we think that this isn't something a user of this package would consider to be a feature.

```suggestion
### Changed
```

cryptodev-2s: Thanks Mark updated

mikesposito (approved):

PR #458, opened 2024-11-08
Standardize tsconfig files

mcmire:
This project has two kinds of TypeScript config files, which is consistent with the module template. However, whereas in the module template, the default config file (`tsconfig.json`) is the one used for development and `tsconfig.build.json` is used to build the project, in this repo it's reversed (`tsconfig.test.json` used for development).

This commit aligns the kinds of TypeScript configuration files with the module template. It also normalizes the existing options to fit the module template. There should be no differences to `dist/`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,10 +25,6 @@ module.exports = {
 
     {
       files: ['*.test.ts', '*.test.js'],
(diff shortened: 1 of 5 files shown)


Gudahtt (approved):
LGTM!

PR #459, opened 2024-11-08
Remove test-helpers from build

mcmire:
The test helpers file is showing up in the published package when it should not be. This commit moves it to a directory that is automatically excluded by the TypeScript configuration.


Gudahtt (dismissed):
LGTM!

Gudahtt:
We need to reduce coverage thresholds to account for removing the test helpers (they were previously considered as code under test)

PR #468, opened 2024-11-26
chore: Fix development `peerDependency` warnings

Gudahtt:
The `@lavamoat/preinstall-always-fail` package has been added, and the package `eslint-plugin-import` package has been downgraded by a minor version, both to meet `peerDependency` requirements of other development dependencies.

The import plugin downgrade resulted in new lint warnings about import order, which were addressed.

Resolves two warnings, no functional changes.

### DIFF
--- a/src/SmartTransactionsController.test.ts
+++ b/src/SmartTransactionsController.test.ts
@@ -17,8 +17,8 @@ import {
 import nock from 'nock';
 import * as sinon from 'sinon';
 
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
+import packageJson from '../package.json';
+import { advanceTime, flushPromises, getFakeProvider } from '../tests/helpers';
 import { API_BASE_URL, SENTINEL_API_BASE_URL_MAP } from './constants';
 import SmartTransactionsController, {
   DEFAULT_INTERVAL,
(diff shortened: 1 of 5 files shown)


Gudahtt:
@SocketSecurity ignore npm/debug@2.6.9

Looks like the `net` package is used for the Socket API, it's not actually making network requests.

Gudahtt:
@SocketSecurity ignore npm/is-async-function@2.0.0

New author is known/trusted

Gudahtt:
@SocketSecurity ignore npm/@lavamoat/preinstall-always-fail@2.1.0

This is our package

src/SmartTransactionsController.test.ts
```diff
@@ -17,8 +17,8 @@ import {
 import nock from 'nock';
 import * as sinon from 'sinon';
 
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
Gudahtt, line 20: It's unclear why these were added, but they are no longer needed

Gudahtt: Oh I see, `yarn tsc` passes but `yarn build` fails due to the tsconfig root directory. I will fix.

Gudahtt: https://github.com/MetaMask/smart-transactions-controller/pull/468/commits/e2cff64674134525147896ee2f0bed6d138ad3e3

Only the reference in production code needed to be ignored. Not the one on this line.

mcmire (approved):
LGTM!

PR #470, opened 2024-11-26
Fix tsconfig to emit JS files in addition to declarations

mcmire:
`tsconfig.build.json` is now instructing TypeScript to only emit type declaration files instead of also emitting JavaScript files. This change was erroneously made in a previous commit. This commit removes that instruction.

## Manual Testing

- Run `yarn build`
- Inspect `dist/`. You should see `.js` and `.js.map` files in addition to `.d.ts` files.


Gudahtt (approved):
LGTM!

PR #471, opened 2024-11-26
chore: Update to NetworkController v22 (and related packages)

Gudahtt:
The `@metamask/network-controller` package has been updated from v21 to v22, and associated dependency bumps required by peer dependency requirements have been bumped as well.

Changelogs:
- `@metamask/network-controller`: https://github.com/MetaMask/core/blob/main/packages/network-controller/CHANGELOG.md#changed-2
  * The only breaking changes are the ones that were mistakenly shipped as v21.1.0, which this package already uses.
- `@metamask/polling-controller`: https://github.com/MetaMask/core/blob/main/packages/polling-controller/CHANGELOG.md#1200
  * Just the network controller peer dependency update
- `@metamask/transaction-controller`: https://github.com/MetaMask/core/blob/main/packages/transaction-controller/CHANGELOG.md#3800
  * Just peer dependency updates


Gudahtt:
We should move these to `peerDependencies` as well, but I'll do that in the next PR

mcmire (approved):
LGTM.

PR #472, opened 2024-11-26
chore: Recategorize controllers as peer dependencies

Gudahtt:
The `@metamask/network-controller` and `@metamask/transaction-controller` dependencies have been moved from `dependencies` to `peerDependencies` (and `devDependencies`). This was done because we're not directly importing and using those controllers directly from imports, they're instead passed in as constructor parameters (indirectly), so it's critical that the versions we use here match the versions used on the client exactly.

The peer dependency requirements of those two packages have been silenced because they aren't actually instantiated here, just used for type purposes (so we don't need their `peerDependencies` present).

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,9 +33,7 @@
     "@metamask/controller-utils": "^11.0.0",
     "@metamask/eth-json-rpc-provider": "^4.1.6",
     "@metamask/eth-query": "^4.0.0",
-    "@metamask/network-controller": "^22.0.0",
     "@metamask/polling-controller": "^12.0.0",
-    "@metamask/transaction-controller": "^38.0.0",
     "bignumber.js": "^9.0.1",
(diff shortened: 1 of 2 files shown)


Gudahtt:
Depends upon #471

package.json
```diff
@@ -74,6 +74,18 @@
     "ts-jest": "^29.1.4",
     "typescript": "~4.8.4"
   },
+  "peerDependencies": {
+    "@metamask/network-controller": "^22.0.0",
+    "@metamask/transaction-controller": "^38.0.0"
+  },
+  "peerDependenciesMeta": {
```
mcmire, line 81: Ah... good idea.

cryptodev-2s (approved):
LGTM!

mcmire (approved):
LGTM.

PR #473, opened 2024-11-26
chore: Remove unnecessary `events` dependency

Gudahtt:
This dependency was not used anywhere. Previously it was brought in because an older version of this controller extended `EventEmitter`, but even then the package wasn't really used (the Node.js events module was resolved instead).

We're not even using `EventEmitter` anymore since the migration to the `BaseControllerV2` base class.


mcmire (approved):
Didn't see this until now! Looks good.

cryptodev-2s (approved):

PR #482, opened 2024-12-12
chore: Update `@metamask/transaction-controller` peer dependency to `42.0.0`

cryptodev-2s:
Update `@metamask/transaction-controller` peer dependency from `38.0.0` to `42.0.0`

Breaking changes introduced in version `41.0.0` of `@metamask/transaction-controller` affect `@metamask/smart-transactions-controller`. These changes involve the removal of global network usage, requiring the explicit configuration of `networkClientId`. [CHANGELOG](https://github.com/MetaMask/core/blob/main/packages/transaction-controller/CHANGELOG.md?plain=1#L71)

### DIFF
--- a/src/SmartTransactionsController.ts
+++ b/src/SmartTransactionsController.ts
@@ -14,6 +14,7 @@ import EthQuery from '@metamask/eth-query';
 import type {
   NetworkClientId,
   NetworkControllerGetNetworkClientByIdAction,
+  NetworkControllerGetStateAction,
   NetworkControllerStateChangeEvent,
 } from '@metamask/network-controller';
 import { StaticIntervalPollingController } from '@metamask/polling-controller';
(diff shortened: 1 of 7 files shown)


src/SmartTransactionsController.ts
```diff
@@ -138,7 +139,9 @@ export type SmartTransactionsControllerGetStateAction =
 export type SmartTransactionsControllerActions =
   SmartTransactionsControllerGetStateAction;
 
-export type AllowedActions = NetworkControllerGetNetworkClientByIdAction;
```
Gudahtt, line 141: Good idea to remove these exports, but we should document this as a breaking change as well

dan437 (approved):
I've tried a send transaction, swap transaction and a dapp transaction with this branch and it worked.

PR #492, opened 2025-01-08
Add request url to exception payload to enable tracking in sentry

antonydenyer:
Add additonal informatinon to errors to help with error tracking

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -200,6 +200,7 @@ export async function handleFetch(request: string, options?: RequestInit) {
     console.log(`response`, response);
     throw new Error(
       `Fetch error:${JSON.stringify({
+        request,
         status: response.status,
         ...mapKeysToCamel(json),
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -200,6 +200,7 @@ export async function handleFetch(request: string, options?: RequestInit) {
     console.log(`response`, response);
     throw new Error(
       `Fetch error:${JSON.stringify({
+        request,
```
Gudahtt, line 203: Dynamic data in error messages has been difficult to work with in the past. It makes it easy to accidentally introduce PII, and it's bad for error grouping on Sentry (each unique message pops up as a separate issue). I'd strongly recommend looking for another solution.

If there is a particular network request / failure case you want more information about, I'd recommend catching there instead and capturing a more detailed exception. This fetch function is used in a variety of places

PR #493, opened 2025-01-24
Fix: SmartTransactionsController `state` should persist

httpJunkie:
feat: Enable Persistence for SmartTransactionsController

Adds state persistence for the SmartTransactionsController, enables access to historical STX data across browser sessions.

Changes:

- Set `persist: true` in `SmartTransactionsController` metadata
- Added `updateState` public method for testing purposes
- Implemented test to verify state persistence between controller instances
- Updated initialization test to confirm persistence is enabled

Benefits
Allows migration scripts and other features to access and utilize historical smart transaction data after browser restarts, improving data continuity and application functionality.

* Fixes #12345
* See: [EXT - #28854](https://github.com/MetaMask/metamask-extension/pull/28854/files#diff-fd9867707661a47627f22edcf23756d1b2bbb0cff370f62e1218d99d8522b6a2R66) and [MOB - #12857](https://github.com/MetaMask/metamask-mobile/pull/12857/files#diff-9a67a9deaa4ac067d33514236de82eec0eef082f3eaee90e3face27fae446d5eR113) features that have migrations script that will be able to use that data.

### DIFF
--- a/src/SmartTransactionsController.test.ts
+++ b/src/SmartTransactionsController.test.ts
@@ -337,8 +337,7 @@ describe('SmartTransactionsController', () => {
     nock.cleanAll();
   });
 
-  // update test name to reflect the new persistence enabled state
-  it('initializes with default state and persistence enabled', async () => {
+  it('initializes with default state', async () => {
     const defaultState = getDefaultSmartTransactionsControllerState();
     await withController(({ controller }) => {
       expect(controller.state).toStrictEqual({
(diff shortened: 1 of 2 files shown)


src/SmartTransactionsController.test.ts
```diff
@@ -349,9 +350,51 @@ describe('SmartTransactionsController', () => {
           },
         },
       });
+      // update to check the persist value
+      expect(controller.metadata.smartTransactionsState.persist).toBe(true);
     });
   });
 
+  it('persists smart transactions state across browser sessions', async () => {
```
cryptodev-2s, line 358: This test is unnecessary because it doesn’t actually verify persists; instead, it focuses on whether the initial state is correctly saved in the Controller State.

src/SmartTransactionsController.test.ts
```diff
@@ -337,7 +337,8 @@ describe('SmartTransactionsController', () => {
     nock.cleanAll();
   });
 
-  it('initializes with default state', async () => {
+  // update test name to reflect the new persistence enabled state
+  it('initializes with default state and persistence enabled', async () => {
```
cryptodev-2s, line 341: There’s no need to update this test either, as handling persistence is not the responsibility of this controller, and we typically don’t verify persistence in our controllers.

cryptodev-2s:
Looks good to me! I’ve left a few comments—let me know your thoughts on addressing them.

src/SmartTransactionsController.test.ts
httpJunkie: I see your point. I will make some adjustments. I appreciate the feedback.

httpJunkie:
> Looks good to me! I’ve left a few comments—let me know your thoughts on addressing them.

Before I make changes, I'm new to working with this package, I know it is used in both Extension and Mobile and I have used it's functionality there. But.. What would you suggest the best tests are considering this change from `false` to `true`.

Do you believe that the addition of this property is fully covered with existing tests and maybe I don't need to update any tests considering the change is just flipping it's value?

Should there be tests anywhere else maybe in the other codebases that use it to account for the change in persistence?

dan437 (approved):

PR #498, opened 2025-02-27
Feat/multichained smart transactions

salimtb:
the goal of this PR is to make the smart transaction multichained and take an array of chainId on the polling input params

### DIFF
--- a/src/SmartTransactionsController.ts
+++ b/src/SmartTransactionsController.ts
@@ -328,9 +328,7 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
     // with a networkClientId that is not supported, but for now I'll add a check in case
     // wondering if we should add some kind of predicate to the polling controller to check whether
     // we should poll or not
-    // const chainId = this.#getChainId({ networkClientId });
-
-    const filteredChainIds = chainIds.filter((chainId) =>
+    const filteredChainIds = (chainIds ?? []).filter((chainId) =>
       this.#supportedChainIds.includes(chainId),
     );
 
@@ -343,10 +341,10 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
   checkPoll({
     smartTransactionsState: { smartTransactions },
   }: SmartTransactionsControllerState) {
-    const allChainsCurrentTransactions =
+    const smartTransactionsForAllChains =
       Object.values(smartTransactions).flat();
-    const currentSmartTransactions = allChainsCurrentTransactions;
-    const pendingTransactions = currentSmartTransactions?.filter(
+
+    const pendingTransactions = smartTransactionsForAllChains?.filter(
       isSmartTransactionPending,
     );
     if (!this.timeoutHandle && pendingTransactions?.length > 0) {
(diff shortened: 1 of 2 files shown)


src/SmartTransactionsController.ts
```diff
@@ -322,23 +322,30 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
-    networkClientId,
+    chainIds,
   }: SmartTransactionsControllerPollingInput): Promise<void> {
     // if this is going to be truly UI driven polling we shouldn't really reach here
     // with a networkClientId that is not supported, but for now I'll add a check in case
     // wondering if we should add some kind of predicate to the polling controller to check whether
     // we should poll or not
-    const chainId = this.#getChainId({ networkClientId });
-    if (!this.#supportedChainIds.includes(chainId)) {
+    // const chainId = this.#getChainId({ networkClientId });
```
dan437, line 331: Can you removed this unused code, please?

src/SmartTransactionsController.ts
```diff
@@ -322,23 +322,30 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
+
+    const filteredChainIds = chainIds.filter((chainId) =>
```
dan437, line 333: Will `chainIds` always be an array? If it could be undefined, this would throw an error `Cannot read properties of undefined (reading 'filter')`

src/SmartTransactionsController.ts
```diff
@@ -322,23 +322,30 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
+    return this.updateSmartTransactions({ chainIds: filteredChainIds });
   }
 
   checkPoll({
     smartTransactionsState: { smartTransactions },
   }: SmartTransactionsControllerState) {
-    const currentSmartTransactions = smartTransactions[this.#chainId];
+    const allChainsCurrentTransactions =
+      Object.values(smartTransactions).flat();
+    const currentSmartTransactions = allChainsCurrentTransactions;
```
dan437, line 348: nit: the idea of having the prefix `current` was meant for the currently selected chain. Since this will be for all chains, can we get rid of this line, do this:
```
const smartTransactionsForAllChains = Object.values(smartTransactions).flat();
```
And use `smartTransactionsForAllChains` later on in the code?

src/SmartTransactionsController.ts
```diff
@@ -589,25 +598,43 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
-      .map((smartTransaction) => smartTransaction.uuid);
+    // Iterate over each chain group directly
+    for (const [chainId, transactions] of Object.entries(smartTransactions)) {
+      if (chainIds && !chainIds.includes(chainId as Hex)) {
+        continue;
+      }
+      // Filter pending transactions and map them to the desired shape
+      const pendingTransactions = transactions
+        .filter(isSmartTransactionPending)
+        .map((tx) => {
```
dan437, line 622: nit: can we rename `tx` to `pendingSmartTransaction`?

src/SmartTransactionsController.ts
```diff
@@ -970,6 +1015,25 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
...
+        this.#supportedChainIds.includes(chainId as Hex),
+    );
+  }
+
+  #getNetworkClientId({ chainId }: { chainId: string }): string {
+    const { networkConfigurationsByChainId } = this.messagingSystem.call(
+      'NetworkController:getState',
+    );
+    return networkConfigurationsByChainId[chainId as Hex].rpcEndpoints[
+      networkConfigurationsByChainId[chainId as Hex].defaultRpcEndpointIndex
```
dan437, line 1033: Are these things always present, e.g. `networkConfigurationsByChainId[chainId as Hex].defaultRpcEndpointIndex`?

src/SmartTransactionsController.ts
```diff
@@ -1044,7 +1108,7 @@ export default class SmartTransactionsController extends StaticIntervalPollingCo
     const {
       smartTransactionsState: { smartTransactions },
     } = this.state;
-    const currentSmartTransactions = smartTransactions?.[this.#chainId];
+    const currentSmartTransactions = Object.values(smartTransactions).flat();
```
dan437, line 1111: nit: Same comment as I mentioned somewhere else, I would rename it to: `smartTransactionsForAllChains` and use that naming later on.

src/SmartTransactionsController.ts
salimtb: yes the value inside networkConfigurationsByChainId are referring to the networkController state and always present

dan437 (approved):

PR #511, opened 2025-04-15
fix: ensure STX status updates across all networks

httpJunkie:
Thanks for reviewing! Here's what you need to know about these changes:

**Current State and Need for Change:**
* Currently, when users have a pending Smart Transaction (STX) and switch to a different network, the STX status fails to update
* Users must switch back to the original network (Ethereum or BNB) to see status updates
* This blocks users from initiating new transactions on other networks until they switch back
* The root cause is that we're using the transaction's chainId instead of the storage key's chainId when fetching status updates

**Solution:**
* Modified `updateSmartTransactions` to consistently use the storage key's chainId instead of the transaction's chainId
* Added comprehensive tests to verify status updates work correctly across multiple networks
* The fix ensures that STX status updates happen regardless of which network the user is currently on
* Users can now freely switch networks without losing track of their pending transactions

**Related Issues:**
* Fixes MetaMask/metamask-extension#31310
* This is the smart-transactions-controller portion of the fix for the extension issue

The changes are minimal but resolve a significant UX pain point where users were forced to switch back to original networks to see transaction status updates.

### DIFF
--- a/src/SmartTransactionsController.test.ts
+++ b/src/SmartTransactionsController.test.ts
@@ -1578,6 +1578,76 @@ describe('SmartTransactionsController', () => {
       );
     });
 
+    it('iterates over transactions from all networks when updating statuses', async () => {
+      // Setup: Create controller with pending transactions on two different networks
+      const mainnetPendingTx = {
+        uuid: 'mainnet-tx',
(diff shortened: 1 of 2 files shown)


httpJunkie:
The issue this fix was for seems to not be an issue anymore.

PR #524, opened 2025-08-08
feat: Get Sentinel URL and supported chain IDs dynamically

dan437:
## Description
This PR allows us to:
- Call the `getSentinelUrl` function with a chainId to get the right Sentinel URL. With this we no longer need to make changes in this codebase when adding a new network.
- Call the `getSupportedChainIds` function, which will return a list of supported chain IDs based on their availability in remote-api-config.


dan437:
Can be reopened when it's prioritised.

PR #535, opened 2025-09-17
Remove default export

mcmire:
We don't use default exports in our libraries. This commit changes the package exports so that `SmartTransactionsController` is a named export.


Gudahtt (dismissed):
LGTM!

PR #541, opened 2025-09-22
fix: export enum values

mikesposito:
Enum were unintentionally exported as types, while client need the values.


Gudahtt (approved):
LGTM!

PR #598, opened 2026-06-12
chore: align external dep ranges with core monorepo

cryptodev-2s:
## Explanation

[`MetaMask/core`](https://github.com/MetaMask/core)'s `yarn.config.cjs` enforces **consistent version ranges for every external dependency across all monorepo packages** (`expectConsistentDependenciesAndDevDependencies`), and `expectUpToDateWorkspaceDependenciesAndDevDependencies` requires internal `@metamask/*` ranges to track the current in-monorepo version. This PR aligns or removes every remaining drift point so the migrated package doesn't fail constraints.

### Aligned

| Dep | Source (before) | Aligned to core |
|---|---|---|
| `@babel/runtime` | `^7.24.1` | `^7.23.9` |
| `@lavamoat/allow-scripts` | `^3.2.1` | `^3.0.4` |
| `typescript-eslint` | `^8.39.0` | `^8.48.0` |
| `@metamask/profile-sync-controller` | `^28.1.1` | `^28.2.0` (core released 28.2.0 since the previous alignment in #590) |

### Removed (unused)

- `@types/node` — no Node API usage anywhere in source (no `fs`/`path`/`crypto`/`process`/`Buffer`; tsconfig `lib` is `["ES2020", "DOM"]`).
- `@types/eslint` — no `import 'eslint'` anywhere.
- `@types/jest` — Jest 29 ships its own types via `@jest/globals`.
- `@types/lodash` — transitively available; `tsconfig.skipLibCheck` makes it irrelevant for source builds.

(Transitive consumers that need any of these install their own copy.)

### Not aligned (intentional)

`nock` stays at `^14.0.0-beta.7`. Downgrading to core's `^13.3.1` causes **22 test failures** (the v14-beta mock interceptors don't match in v13). Two options for Phase B:

- Add `nock` to `ALLOWED_INCONSISTENT_DEPENDENCIES` in core's `yarn.config.cjs`, OR
- Bump core's nock from v13 to v14 separately

### Source-only deps (not in core, will migrate as-is)

`@ethersproject/bytes`, `@ethersproject/keccak256`, `@ethersproject/transactions` (runtime), `@arethetypeswrong/cli`, `prettier-plugin-packagejson` (dev — Phase B PR #10 will strip the dev ones).

### Verification

- ✅ `yarn build`
- ✅ `yarn lint`
- ✅ `yarn jest` — 196/196 pass

## References

- Constraint rules: [`yarn.config.cjs:222 expectUpToDateWorkspaceDependenciesAndDevDependencies`](https://github.com/MetaMask/core/blob/main/yarn.config.cjs#L222) and [`yarn.config.cjs:885 expectConsistentDependenciesAndDevDependencies`](https://github.com/MetaMask/core/blob/main/yarn.config.cjs#L885)


package.json
```diff
@@ -66,7 +66,7 @@
   },
   "devDependencies": {
     "@arethetypeswrong/cli": "^0.18.2",
-    "@lavamoat/allow-scripts": "^3.2.1",
+    "@lavamoat/allow-scripts": "^3.0.4",
```
cryptodev-2s, line 69: I could also update in core if it's better

package.json
```diff
@@ -41,7 +41,7 @@
     "test:watch": "jest --watchAll"
   },
   "dependencies": {
-    "@babel/runtime": "^7.24.1",
```
cryptodev-2s, line 44: I could also update in core if it`s better

package.json
Mrtenz: We can do that after the merge. This is fine for now.

Mrtenz (approved):
Please review the Socket report before merging.

cryptodev-2s:
@SocketSecurity ignore npm/glob@13.0.6

False positive. The "fetch" identifiers Socket detected come from `lru-cache` (bundled into glob 13's minified output), not HTTP fetch. They are LRUCache's in-memory cache loader API (`fetchMethod`, `fetchAborted`, `fetchDispatched`, etc.). Transitive devDep of `@lavamoat/allow-scripts`; no HTTP access occurs.

cryptodev-2s:
@SocketSecurity ignore npm/nock@14.0.15

False positive. nock's source (`lib/intercept.js`, `index.js`, etc.) is unobfuscated, readable JavaScript with normal identifiers and JSDoc. The signal is likely triggered by minified bundles. Industry-standard HTTP mocking library used here as a devDep for tests only.

cryptodev-2s:
@SocketSecurity ignore npm/@mswjs/interceptors@0.41.9

False positive. The "obfuscation" detection is triggered by hash-suffixed chunk filenames (e.g. `fetchUtils-umV5xXBy.cjs`, `BatchInterceptor-BGhH8z1z.cjs`) that modern bundlers produce for cache busting. Chunk contents and `src/` TypeScript files are readable with proper identifiers. Transitive devDep of `nock@14`; used only at test time.

cryptodev-2s:
@SocketSecurity ignore npm/undici@6.26.0

All 5 alerts checked, all false positives.

- **Network access in `globalThis.fetch`**: undici *is* Node.js's official `fetch` implementation (powers `globalThis.fetch` since Node 18). HTTP access is its purpose. Maintained by Matteo Collina (Node.js TSC). Homepage: undici.nodejs.org.
- **AI anomaly (UTF-8 → Latin-1 re-encoding)**: This is `scripts/strip-comments.js`, a build-time script that transcodes undici's own browser bundle (`undici-fetch.js`) for CDN/bundler compatibility. Not runtime code; operates only on undici's own internal file at build time.
- **AI anomaly (cache batch operations)**: HTTP response caching. The Socket AI's own notes state: *"No signs of malware, data exfiltration, backdoors, or obfuscated behavior were found."*
- **AI anomaly (HTTP/WebSocket upgrade handler)**: Standard protocol handling. The Socket AI's own notes state: *"no indicators of malware or obfuscation are detected."*
- **AI anomaly (4th)**: Same class of HTTP-client behaviour signal as the others.

Transitive devDep of `@lavamoat/allow-scripts` (build/install tooling). Not used at runtime by this package.

